PR #89, opened 2025-07-01
88 render markdown tables in chat

JulienOrnat:


frontend/src/components/chatbot/MessageCard.tsx
```diff
@@ -135,13 +135,29 @@ export default function Message({
...
+                    },
+                    "& th, & td": {
+                      border: "1px solid",
+                      padding: theme.spacing(1),
+                      textAlign: "left",
+                    },
+                    "& th": {
+                      backgroundColor: theme.palette.grey[100],
+                      fontWeight: "bold",
+                    },
```
florian-muller, line 160: This mardown table css here will only be used  by this instance of  `MarkdownRenderer` and not other, like in `<DocumentViewer />` :

image

To keep the same logic, I think we should update the `getMarkdownComponents` function used in `MarkdownRenderer`. Try to use MUI tables components (or mui styles at least) to keep a coherent appearance.

It should help supporting dark themes which does not look great for now:

image

florian-muller (changes requested):

PR #171, opened 2025-07-16
because I need a test for pr testing

cbid71:


cbid71:
Tests on PR events here.
OK.

PR #179, opened 2025-07-17
#178 add missing apache headers

dimitri-tombroff:


dimitri-tombroff:
only added apache headers wherever missing

PR #210, opened 2025-07-23
206 add dynamic agent creation

### LINKED ISSUE #206: Add dynamic agent creation

simcariou:


agentic_backend/app/features/dynamic_agent/service.py
```diff
@@ -3,23 +3,41 @@
 from app.chatbot.dynamic_agent_manager import DynamicAgentManager
 from app.features.dynamic_agent.mcp_agent import MCPAgent
 from app.features.dynamic_agent.structures import MCPAgentRequest
+from app.application_context import get_app_context, get_configuration
+from app.common.structures import AgentSettings
+
+class DynamicAgentManagerService:
```
florian-muller, line 9: It seems there is 2 way 2 create a dynamic agent that are redundant:

- with `get_create_agent_instance` from `AgentManager` (that will do its own caching)
- with `get_create_agent_instance` from `DynamicAgentManager` (that does too its own caching)

I don't see the interest of having this 2 very similar method (but maybe I am missing something, sorry it its the case).

If they really are the same, I think we should only keep the one from `AgentManager`, it allow us to have a simple unique function to get an agent, independently of its type (dynamic or not) and to handle agent caching in one place.

florian-muller, line 9: (could not comment on the good line)

Seems like the `clear_cache_for_session` of `DynamicAgentManager` is not used ? It it's the case, I think we should remove it

agentic_backend/app/chatbot/dynamic_agent_manager.py
florian-muller, line 1: (could not comment on line, again)

suggestion: `get_agentic_flows` from `DynamicAgentManager` is called directly by the a controller. I think it could be nice to call it in `get_agentic_flows` of `AgentManager` to also keep one function to get all agentic flows

agentic_backend/app/chatbot/agent_manager.py
```diff
@@ -69,8 +71,13 @@ def get_create_agent_instance(self, name: str,
...
             self.agent_cache[cache_key] = agent_instance
             return agent_instance
-
-        agent_instance = agent_class(cluster_fullname=argument)
+        else:
+            if self.dynamic_agent_manager.has_agent(name):
+                agent_instance = self.dynamic_agent_manager.get_agent_instance(name)
+            else:
+                agent_class = get_agent_class(name)
+                agent_instance = agent_class(cluster_fullname=argument)
```
florian-muller, line 79: As I understand it, dynamic agent class are only created in `get_create_agent_instance` of `AgentManager` and we can see here that it use `self.dynamic_agent_manager.get_agent_instance` and never `get_agent_class(name)` so: why do we update `agent_classes` in `ApplicationContext` with dynamic agent classes if it will never be used ?

I also have an intuition that using `get_agent_class(name)` on a dynamic agent would not work as here, the `agent_class` only take a `cluster_fullname` and our dynamic agent classes (`MCPAgent`) need more.

I think it create again 2 way to get a dynamic agent instance or class and that create confusion. If that is really the case, we should either:

- keep `dynamic_agent_manager.get_agent_instance` but not update `agent_classes` in `ApplicationContext`
- keep the actual logic, make `get_agent_class(name)` work with dynamic agent, register them in `agent_classes` in `ApplicationContext` and remove `get_agent_instance` from `ApplicationContext`

Not sure which one is the best

florian-muller (dismissed):
Great work to have implemented this so quickly. It's easy to get lost in all these agent instantiation logic (I got a bit lost while reviewing). 

When reviewing quickly, I found the agent creating a bit complex (need to register new agent in lots of places: `_agent_index`, `_agent_index` and `agent_store`) but I had not time to read more and see it could be simplified. Maybe can talk about it some day :)

agentic_backend/app/features/dynamic_agent/service.py
```diff
@@ -28,12 +46,12 @@ def constructor() -> MCPAgent:
...
 
-    dynamic_agent_manager.register_agent(req.name, constructor)
-    return {"status": "success", "agent_name": req.name}
+        def constructor() -> MCPAgent:
+            return agent_instance
+
+        self.dynamic_agent_manager.register_agent(req.name, constructor, MCPAgent)
 
+        agent_store = get_app_context().get_dynamic_agent_store()
+        agent_store.save(agent_instance)
```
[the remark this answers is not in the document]
simcariou: - [] Yes you are right thx

agentic_backend/app/features/dynamic_agent/stores/mcp_agent/duckdb_mcp_agent_store.py
```diff
@@ -0,0 +1,106 @@
...
+            role=row[2],
+            nickname=row[3],
+            description=row[4],
+            icon=row[5],
+            categories=json.loads(row[6]) if row[6] else [],
+            tag=row[7],
+            cluster_fullname="", # Placeholder waiting to get rid of it as a non optional param in Agen8tFlow
+        )
+
+    def save(self, agent: AgentFlow) -> None:
```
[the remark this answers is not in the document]
simcariou: [ ] Done j'ai mis AgentFlow pour rester carré

agentic_backend/app/features/dynamic_agent/service.py
simcariou: - [x] Done

agentic_backend/app/features/dynamic_agent/service.py
```diff
@@ -3,23 +3,41 @@
...
+        Returns the dynamic agent manager for integration elsewhere
+        """
+        return self.dynamic_agent_manager
+
+    def build_and_register_mcp_agent(self, req: MCPAgentRequest):
+        """
+        Builds, registers, and stores the MCP agent, including updating app context and saving to DuckDB.
+        """
+        if req.name in self.dynamic_agent_manager.get_registered_names():
+            raise HTTPException(status_code=409, detail=f"Agent '{req.name}' already exists")
```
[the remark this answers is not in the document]
simcariou: - [x] Très bien vu j'ai corrigé

agentic_backend/app/common/structures.py
```diff
@@ -255,6 +255,22 @@ class OpenSearchStorageConfig(BaseModel):
...
 ]
 
+###########################################################
+#
+#  --- Dynamic Agents Storage Configuration
+#
+
+class LocalDynamicAgentStorage(BaseModel):
+    type: Literal["local"]
+    local_path: str = Field(default=str(Path("~/.fred/dynamic_agents.json")), description="Local storage json file")
```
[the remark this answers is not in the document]
simcariou: Non tu as raison, j'enlève

agentic_backend/app/features/dynamic_agent/stores/mcp_agent/duckdb_mcp_agent_store.py
```diff
@@ -0,0 +1,106 @@
...
+    def load_all(self) -> List[AgentFlow]:
+        with self.store._connect() as conn:
+            rows = conn.execute(
+                f"SELECT * FROM {self.store._prefixed(self.table_name)}"
+            ).fetchall()
+        loaded_agents = [self._deserialize(row) for row in rows]
+        logger.info(f"Fetched all dynamic agents: {loaded_agents}")
+        return loaded_agents
+
+    def delete(self, name: str) -> None:
```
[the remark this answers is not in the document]
simcariou: - [x] Done

agentic_backend/app/tests/conftest.py
```diff
@@ -88,6 +88,7 @@ def minimal_generalist_config() -> Configuration:
         tool_metrics_storage={"type": "local", "local_path": "/tmp/tool-metrics"},
         feedback_storage={"type": "local", "local_path": "/tmp/feedback"},
         session_storage={"type": "in_memory"},
+        dynamic_agent_storage={"type": "duckdb", "duckdb_path": "~/.fred/db.duckdb"},
```
simcariou, line 91:
```suggestion
        dynamic_agent_storage={"type": "duckdb", "duckdb_path": "/tmp/db.duckdb"},
```

PR #233, opened 2025-07-28
Test-openapi-drift-check-ci

florian-muller:


florian-muller:
Was only created to see if CI failed

PR #234, opened 2025-07-29
227 start ordering must be removed between agentic and knowledge flow

dimitri-tombroff:
i added a loop to restart the agent checking so that start ordering is ok. Will create a ticket to handle the more complex case where knowledge flow disappear then comes back later on.


dimitri-tombroff:
Pour le moment on vérifie toutes les 10 secondes, les agents repassent dans un dictionaire ok ou restent dans un dictionaire failed. C'est assez simple.

PR #240, opened 2025-07-29
237 improve fred for the bid use case

dimitri-tombroff:


dimitri-tombroff:
I am going fast for this one to unlock the bid experiment. To improve in the future: centralize model and embedding settings in the fred-core library.

PR #261, opened 2025-07-31
260 revisit frontend security activation

cbid71:


florian-muller (changes requested):
I don't think the chosen method to communicate the `USE_AUTH` is optimal. Their is already 2 ways to communicate settings handled by `loadConfig`:
- the `front/public/config.json` file
- the `/agentic/v1/config/frontend_settings` route returning the `frontend_settings` of the `configuration.yaml` (which already has a `feature_flags` list)

I don't think adding a third way just to communicate this info is good.

Moreover, I think using a `VITE_` variable is confusing as Vite env variable (always starting with `VITE_`) are not supposed to be editable at image start time.

----

The solution I would chose would be to add the info in the `/agentic/v1/config/frontend_settings` route, based on the `security.enabled` property in `configuration.yam`.

If authent is enabled in the back, it is enabled in the front automaticaly. 

Here is a minimal example of how I would change the controller:

```python
class FrontendSettings(BaseModel):
    feature_flags: FrontendFlags
    properties: Properties


class TotalFrontendSettings(FrontendSettings):
    security: bool

@app.get(
    "/config/frontend_settings",
    summary="Get the frontend dynamic configuration",
    tags=fastapi_tags,
)
def get_frontend_config():
    conf = get_configuration()
    return TotalFrontendSettings(
        feature_flags=conf.frontend_settings.feature_flags,
        properties=conf.frontend_settings.properties,
        security=conf.security.enabled,
    )
```

PR #269, opened 2025-08-04
266 add prompt support

dimitri-tombroff:


dimitri-tombroff:
I merge this because it fixes many important issues in particular the limit in documents (10) when used using opensearch. Plus a number of improvment and simplification to deal with the proper delete of documents metadata.

PR #289, opened 2025-08-06
273 prepare a first stable release

dimitri-tombroff:


dimitri-tombroff:
i added the missing store i continued to simplify the code and fix lint or type errors

PR #292, opened 2025-08-06
290 repair the monitoring UI page

dimitri-tombroff:


dimitri-tombroff:
ouf ! the session store must be separated int two smaller but we will do that next

PR #486, opened 2025-09-12
add profile to resources and profile selector

JulienOrnat:


albancapitant-thales:
What about using a radio button instead of a square button to select the profile? 

I think it would be clearer for the user that the user can only select one profile (which makes sense).

albancapitant-thales:
Hello, 

J'ai regardé ta PR : 

- Fonctionnellement, ça m'a l'air Ok (j'ai fait quelques tests). Seul petit point : Utiliser un radio button plutôt qu'un squared button serait un peu plus intuitif pour l'utlisateur à mon avis !
- Il semble qu'il y aie des jobs de CI qui failent, à checker ! 
- J'ai essayé de rebase ta branche sur la dernière version de main, et j'ai l'impression que je me tape beaucoup de conflits, je ne sais pas si c'est normal, à checker ! 
- Sinon, sur le code front en tant que tel, je n'ai rien noté de spécial. Mais je suis loin d'être un expert. Donc ça vaut peut-être le coup que tu demandes une review rapidos par Florian ou Dimitri ?

Une fois que tout ça est Ok, go merge !

florian-muller:
Hey hey,

I haven't read the code for now, just checked out the branch and tested the features. I tested it using the prod environment (`make run-prod`).

I noticed a few bugs and had a few improvements idea, here they are:

## Bugs
### Profile managing

- [ ] When creating a Profile, you get a success notification but the new profile doesn't appear in the list, you have to refresh the page to see it. This must be because of OpenSearch (it acknowledge the change but only really update its shards in batch every 1s: if you send the updated and refresh the list before the update batch, you will get the old list... yeah...). A quick fix I found is using `"refresh": "wait_for"` param, the opensearch request will wait for the refresh before returning (so it can take 1 whole seconds... yeah...). I did it in the tag store, [here](https://github.com/ThalesGroup/fred/blob/a20cf901f0de26eb2b8bc68ebc3215db207e11b4/knowledge_flow_backend/app/core/stores/tags/opensearch_tags_store.py#L96)

- [ ] When creating 2 Profiles in a row (without changing page/refreshing), when opening the form for the second time, it is still filled with the values of the previously created Profile.

- [ ] When editing a profile, the form is empty. I think it should be filled with the old value to be able to edit it (like just add a new sentence, not retype the whole thing). 

-  [ ] When deleting a Profile or a Profile Library (Tag), I get a 500 `Unsupported tag type: TagType.PROFILE`. Seems like the tag service must be updated to support the profiles ? (not sure, I haven't look at the new implementation).
  ```
  ERROR    2025-09-19 17:14:18 - ERROR - fastapi_handlers.py - Unhandled exception in PUT                                                         
           http://localhost:8111/knowledge-flow/v1/tags/eb99b7c3-7005-4013-a671-74586fcf8b71: Unsupported tag type: TagType.PROFILE 
  ``` 
## Improvements

### Profile selector (on chat page)

- [ ]  The selected Profile should be remembered. Here if you change page/close and re-open the page, you have to re-select your Profile every time ! Which take at least 3 clicks and could require scroll or typing. In my opinion, n one is going to use this feature in this condition (or it will be frustrating when someone forget to reset it's Profile for then 10th time of the day).
Like what we have for the agent selection, the last Profile used by conversation should be stored somewhere (like the browser local storage). A better way could be to get the last used Profile from the chat history (if it's stored in it). Even for new chat, the last Profile used for the last new chat should be selected by default.

- [ ] The selector does not handle "sub-libraries" (sub-directories ?), all the Profile libraries are at the root level, this is really confusing. If you have 2 sub-librairies with the same name, you can't distinguish them. 

- [ ] The search bar only search for the library name, that doesn't seem that helpful ? In addition, the search bar on the  "knowledge hub" page search only for the Profiles names and not the librairies. It feels weird to have 2 different searches for the same resource.

- [ ] I think the libraries should be "opened" (with their profiles shown under them) by default

- [ ] Empty libraries could be masked here ? As they can't be used


------

As I said, I haven't look at the code, so there may be bugs not coming from your changes. I leave it to you to see which ones should be fixed in this PR and which are out of the scopes and should have their own issue. 

And do not hesitate to contact me if you want to talk about how to address some of this point :)

PR #556, opened 2025-09-24
500 improved UI representation of agent capabilities and plugins

JulienOrnat:


florian-muller (commented):
Nice feature ! I like the Popover design : ) , it's a good thing to bring information and transparency to our users

## Testing suggestions

I have just tested the feature for now (not read the code), here are some suggestions/idea I had: 

- [x] "Search" value should be aligned right like the other:
  Align left:
  Image

  Align right:
  Image

- [x] Small suggestion: as all the tokens line are in a "group", I think the spacing between the "Token used" and the 2 "From..." lines should be smaller than the default gap between 2 lines (8px).

  With 8px top margin:
  Image

  With no margin:
  Image

- [x] As I understand the "Profile" feature, you can only have one "profile" enabled at a time ? If that's the case, it would make sense to me to write "Profile" instead of "Profiles" and not display it as an information that could be a list (like librairies).

  suggestions:
  Image

  Image

- [x] We could try to put more information on the librairies and profile use indicator ? For the profile at least, we could indicate it's name ? (with a max length and `...` if it's too long)

  Image

  Not sure, what do you think about it ?

- [x] We could also show more information in the tooltips, like list of libraries and profile used ?

  Image

  _If we already display the profile directly it is maybe a bit redundant, but it could be nice for the libraries_

- [x] When there is no libraries and / or no profile use, we could maybe just not show the indicator ? Not clutter the screen ?

- [ ] If we write "AI content may be incorrect" in the Popover, we could maybe remove it from under the message ? 

- [ ] Same for the token count, it is more detailed in the Popover, I think we should remove it from under the message

- [ ] If you are ok with the 2 last point, as we have remove lots of redundant information under the message, we could maybe move the libraries/Profile and metadata indicator under the message ? It would maybe help not having indicator all over the place ?

  on top
  Image
  under:
  Image

  Image

- [ ] I like the fact that those information only appear on hover because it de clutter the interface. I just fear that sometimes user might not know its there. I looked a bit at what others are doing for inspiration. In "Le Chat" (mistral), the "bottom metadata" of the **last message** are always shown. For the other messages, you have to hover the message to see them:
  Image
  _Here I was hovering the lightning under the first message_

  It could be a good idea, not sure. 

All these are suggestions, I leave it to you to see the one you like : )

florian-muller (changes requested):

PR #565, opened 2025-09-26
564 provide an integrated log viewer to facilitate production maintainance

dimitri-tombroff:


dimitri-tombroff:
I push this because very useful for devs or production. Including a copy all and copy button to easily send info to support.

PR #580, opened 2025-10-01
577 able to scroll entire fred chat page

simcariou:


florian-muller (approved):
Nice fix. 

In future, it would be nice to not limit that much the documents card size because, on big screen, now there is lot of wasted space:

Image

But that's better than the bugged scroll we had before :)

PR #598, opened 2025-10-03
594 context   the context feature is not working

thomas-hedan:
**First problem:** Only the generalist agent had access to the profiles.

**Second problem:** The profile description was injected at the beginning of the conversation, so the LLM couldn’t reference it once the conversation was already underway.

Solution : Added the code to every agents and the profile description is added at the end of the messages so the LLM cant forget about it.

Tested on agents : George, Tessa, Bronte and leader (Not working so much)


agentic_backend/app/agents/content_generator/content_generator_expert.py
```diff
@@ -148,8 +148,7 @@ async def _reasoner(self, state: MessagesState):
 
         # 4. Wrap and invoke the model.
         messages = self.with_system(sys_prompt, state["messages"])
-
-        assert self.model is not None
```
albancapitant-thales, line 152: Why did you delete this line?

thomas-hedan: Yes it's not present in other agents that are working just fine.

dimitri-tombroff (approved):

Issue #919, opened 2025-12-19
sifferent users cannot create use the same library name


dimitri-tombroff:
for example 'personal'. This is a simple opensearch issue. Not a serious id issue

PR #920, opened 2025-12-19
#919 allow user to create their own libraries without conflict

dimitri-tombroff:


dimitri-tombroff:
tested on a poruction setup. I performed a complete test that has been used for capturing the blog screenshots

PR #1101, opened 2026-02-06
feat: allow tags and document to be owned by team

florian-muller:
## Changes

backend:
- create tag service now take an option `team_id` parameter. If set (and use has rights), tag is creaed in the team instead of in personal user space
- list tag service can be filtered to list personal or team tag
- list tag service return tag permission for each tag
- Fix postgres tag store that was filtering tags by owner id stored in postgres instead of letting ReBAC handle it

frontend:
- `DocumentLibraryList` can now take an optional `teamId` to display tag of a team. If not provided, it display personal tags
- all RBAC check in `DocumentLibraryList` have been removed and replace by the ReBAC permission system

other:
- Claude Code skill to help debug openfga
- run CI on all PR (not just on PR to `main`)

image

### DIFF
--- a/agentic-backend/config/.env.template
+++ b/agentic-backend/config/.env.template
@@ -49,7 +49,7 @@ OPENAI_API_KEY=""
 # -----------------------------------------------------------------------------
 
 # Postgres credentials
-POSTGRES_PASSWORD=""
+FRED_POSTGRES_PASSWORD=""
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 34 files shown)


florian-muller:
@dimitri-tombroff @simcariou  PR is ready to be reviewed/merge. I kept it in draft because I would prefer to not "squash and merge" (and loose the commit history). If that's good for you, I will merge it by hand in `add-team-page`

dimitri-tombroff (approved):

PR #1222, opened 2026-02-24
Feat: 1180 [UI] Implement new navigation without administration pages

Felix-Page:


frontend/src/common/router.tsx
```diff
@@ -19,32 +19,42 @@ import { ProtectedRoute } from "../components/ProtectedRoute";
...
 import { TeamsPage } from "../pages/TeamsPage.tsx";
 import { getConfig } from "./config";
+import DesignSystemPage from "../pages/DesignSystemPage/DesignSystemPage.tsx";
+import MainLayout from "@shared/layouts/MainLayout/MainLayout.tsx";
+import React, { lazy, Suspense } from "react";
+import LoadingWithProgress from "../components/LoadingWithProgress";
 
 const basename = getConfig().frontend_basename;
 
 const RootLayout = ({ children }: React.PropsWithChildren<{}>) => <LayoutWithSidebar>{children}</LayoutWithSidebar>;
```
florian-muller, line 38: Why keep RootLayout only for the 404 page ? 
It put back the old side bar just for this page:

Image

I think it should use `MainLayout` like the other page. We should remove RootLayout, LayoutWithSidebar and the old sidebar

frontend/src/common/router.tsx
```diff
@@ -19,32 +19,42 @@ import { ProtectedRoute } from "../components/ProtectedRoute";
...
 import { Profile } from "../pages/Profile";
-import RebacBackfill from "../pages/RebacBackfill";
-import Runtime from "../pages/Runtime";
 import { TeamDetailsPage } from "../pages/TeamDetailsPage.tsx";
 import { TeamsPage } from "../pages/TeamsPage.tsx";
 import { getConfig } from "./config";
+import DesignSystemPage from "../pages/DesignSystemPage/DesignSystemPage.tsx";
+import MainLayout from "@shared/layouts/MainLayout/MainLayout.tsx";
+import React, { lazy, Suspense } from "react";
+import LoadingWithProgress from "../components/LoadingWithProgress";
```
florian-muller, line 34: Before, as we were all on vscode, I added in `frontend/.vscode/settings.json`

```
"editor.codeActionsOnSave": {
    "source.organizeImports": "always"
  },
```
to sort import on save. 

To support more IDE and still have imported sorted, It think we should add a prettier plugin to do this (or an other tool if you know one ?).

frontend/src/components/agentHub/A2aCardDialog.tsx
```diff
@@ -71,7 +61,12 @@ export const A2aCardDialog = ({ open, onClose, card }: Props) => {
             <Typography variant="body2" color="text.secondary" sx={{ mt: 0.5 }}>
               {card.description}
             </Typography>
-            <Stack direction="row" spacing={1} sx={{ mt: 1, flexWrap: "wrap" }} divider={<Divider orientation="vertical" flexItem />}>
+            <Stack
+              direction="row"
+              spacing={1}
+              sx={{ mt: 1, flexWrap: "wrap" }}
+              divider={<Divider orientation="vertical" flexItem />}
+            >
```
florian-muller: (Next time, it would be nice to run the formater on a distinct branch/PR to make the review easier and smaller)

frontend/src/components/chatbot/HitlDialog.tsx
florian-muller, line 1: Looks like this file is empty, should we keep it ?

frontend/src/locales/en/translation.json
```diff
@@ -1301,5 +1301,25 @@
   "confirmationDialog": {
     "defaultCancelButtonLabel": "Cancel",
     "defaultConfirmButtonLabel": "Confirm"
+  },
+  "rework": {
+    "sidebar": {
+      "title": "FRED.AI",
+      "team": {
+        "avatarAlt": "Team {{teamName}}",
+        "bannerAlt": "Team's banner",
```
florian-muller, line 1317: So avatarAlt and bannerAlt can be coherent, shouldn't both indicate the team name and what the image is (avatar or banner)

```suggestion
        "avatarAlt": "{{teamName}} team avatar",
        "bannerAlt": "{{teamName}} team banner",
```

frontend/src/locales/en/translation.json
```diff
@@ -1301,5 +1301,25 @@
   "confirmationDialog": {
     "defaultCancelButtonLabel": "Cancel",
     "defaultConfirmButtonLabel": "Confirm"
+  },
+  "rework": {
```
florian-muller, line 1312: Why put this in a `rework` section and not just replace the current `sidebar` section ?

frontend/src/locales/fr/translation.json
```diff
@@ -1302,5 +1302,25 @@
   "confirmationDialog": {
     "defaultCancelButtonLabel": "Annuler",
     "defaultConfirmButtonLabel": "Confirmer"
+  },
+  "rework": {
+    "sidebar": {
+      "title": "FRED.AI",
+      "team": {
+        "avatarAlt": "Équipe {{teamName}}",
+        "bannerAlt": "Bannière de l'équipe",
```
florian-muller, line 1318: (Same here)

```suggestion
        "avatarAlt": "Avatare de l'équipe {{teamName}}",
        "bannerAlt": "Bannière de l'équipe {{teamName}}",
```

frontend/src/rework/components/shared/utils/Types.ts
```diff
@@ -0,0 +1,13 @@
+export type Types = "primary" | "secondary" | "tertiary" | "error" | "success" | "warning" | "info";
```
florian-muller, line 1: As this type is only used for colors, it could maybe be named with a color related name ? Something like `ColorVariants`, `ThemeColors`...

frontend/Makefile
```diff
@@ -79,13 +79,17 @@ format: ## Use prettier to format all codebase
 
 .PHONY: update-agentic-api
 update-agentic-api: ## update the generated hooks used in UI based on agentic backend openApi spec
+	@echo "🔧 Installing frontend dependencies..."
+	cd ../frontend && npm ci
```
florian-muller, line 83: No need to cd, we are already in frontend directory

```suggestion
	npm ci
```

frontend/Makefile
```diff
@@ -79,13 +79,17 @@ format: ## Use prettier to format all codebase
...
 	@echo "🔧 Generating agentic backend OpenAPI specification..."
 	cd ../agentic-backend && make generate-openapi
 	@echo "🎯 Generating RTK Query hooks for agentic backend..."
 	npx -y @rtk-query/codegen-openapi src/slices/agentic/agenticOpenApiConfig.json
 
 .PHONY: update-knowledge-flow-api
 update-knowledge-flow-api: ## update the generated hooks used in UI based on knowledge-flow backend openApi spec
+	@echo "🔧 Installing frontend dependencies..."
+	cd ../frontend && npm ci
```
florian-muller, line 92: same

```suggestion
	npm ci
```

frontend/index.html
```diff
@@ -4,6 +4,11 @@
     <meta charset="UTF-8" />
     <link id="favicon" rel="icon" type="image/svg" href="/images/rift.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link href="https://fonts.googleapis.com/css2?family=Geist:wght@100..900&display=swap" rel="stylesheet" />
```
florian-muller, line 9: I am not sure this font import will work in an air gap environment

frontend/package.json
```diff
@@ -16,33 +22,19 @@
...
     "echarts": "^5.5.1",
     "echarts-for-react": "^3.0.2",
-    "eslint": "^10.0.1",
     "http-proxy-middleware": "^3.0.2",
     "i18next": "^25.2.1",
     "i18next-browser-languagedetector": "^8.2.0",
     "katex": "^0.16.23",
     "keycloak-js": "^25.0.1",
     "leaflet": "^1.9.4",
-    "lodash": "^4.17.23",
```
florian-muller, line 35: (Nice catch :), we're better without all this (especialy lodash))

frontend/src/rework/components/shared/atoms/Button/Button.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface ButtonProps extends ComponentPropsWithoutRef<"button"> {
+  children: React.ReactNode;
+  color: Types;
+  variant: ButtonVariant;
+  size: ButtonSize;
+  icon?: IconProps;
+}
+export default function Button({ children, color, variant, size, icon, className, ...props }: ButtonProps) {
+  const buttonClasses = [styles.btn, styles[`btn-${color}`], styles[`btn-${size}`], styles[`btn-${variant}`]];
```
florian-muller, line 14: must say, I am not a fan of  the class inferance without type safety (we're not sure that class `btn-${color}` exist and would have no error if not). We could maybe add `typed-css-modules` (never tested) to generate type for the css and validate that the generated class name will be valid ?

frontend/src/rework/components/shared/organisms/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,25 @@
...
+import UserProfile from "@shared/molecules/UserProfile/UserProfile.tsx";
+import { useLocation } from "react-router-dom";
+
+export default function Sidebar() {
+  const { pathname } = useLocation();
+
+  const smallNavbar = pathname.startsWith("/teams");
+
+  return (
+    <div className={`${styles["sidebar-container"]} ${smallNavbar ? styles["small"] : ""}`}>
```
florian-muller, line 13: I find this notation a bit convoluted, we could use clsx

It think that could be a bit more readable
```tsx
<div className={clsx(styles["sidebar-container"], smallNavbar && styles["small"])}>
```

frontend/src/rework/components/shared/organisms/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,19 @@
...
+.team-selection-container {
+  grid-row: 1 / span 2;
+}
+
+.user-profile-container {
+  position: absolute;
+  bottom: 0;
+  padding: 0 var(--spacing-xs) var(--spacing-xs);
+  width: 100%;
+}
```
florian-muller, line 19: Don't think this file needs to be scss ?

frontend/src/rework/components/shared/organisms/Sidebar/Sidebar.tsx
florian-muller: Ho, also looks like `small` class does not exist ?

Maybe that's why on full teams page sidebar-container take a lot of space for nothing ?

Image

frontend/src/rework/components/shared/organisms/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,25 @@
+import TeamSelectionNavbar from "@shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx";
+import TeamContentNavbar from "@shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx";
+import styles from "./Sidebar.module.scss";
+import UserProfile from "@shared/molecules/UserProfile/UserProfile.tsx";
+import { useLocation } from "react-router-dom";
+
+export default function Sidebar() {
+  const { pathname } = useLocation();
+
+  const smallNavbar = pathname.startsWith("/teams");
```
florian-muller, line 10: Instead of a hardcoded check here, we could maybe handle the "double sidebar" with 2 "layout" component:

- `LayoutWithTeamSidebar`: put the `TeamSelectionNavbar` on the left and leave the right space for the content
- `LayoutWithFullSidebar`: use `LayoutWithTeamSidebar` and inside, put the `TeamContentNavbar` on the left and leav the right space for the conent

Then I think we could choose what sidebar a route must use more explicitly in rooter.tsx with:
```ts
export const routes: RouteObject[] = [
    {
        path: "/",
        element: <LayoutWithFullSidebar />,
        children: [
            // All classic roots here
        ]
    },
    // Execeptional element with only one sidebar
    {
        path: "teams",
        element: (
            <LayoutWithTeamSidebar>
                <TeamsPage />
            </LayoutWithTeamSidebar>
        ),
    }
];
```

I leave it to you to see if the change is worth :)

frontend/src/common/router.tsx
```diff
@@ -54,6 +64,10 @@ export const routes: RouteObject[] = [
         path: "/new-chat",
         element: <NewChatAgentSelection />,
```
florian-muller, line 65: You may be planning on doing it, but just in case, if I understand well the new UI, we will have to change this route and scope it by team ? like `team/<id>/new-chat` (to get the team id in NewChatAgentSelection and only display the list of team's agent).

Now that I think of it, we should do the same for `/chat/<id>` no ? It would be simpler, just to decide which team should selected in the sidebar

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles["team-navbar-container"]}>
+      <div>
+        <span className={styles.title}>{t("rework.sidebar.title")}</span>
+        <TeamSelectionItem
+          redirection={"agents"}
+          teamName={t("rework.sidebar.team.userTeam")}
+          selected={pathname.startsWith(`/agents`) || pathname.startsWith(`/knowledge`)}
```
florian-muller, line 20: To prepare for "personal space is just a team" we could change the router to move this pages to `/team/personal/agents` and `/team/personal/knowledge` where in the UI we handle `personal` id ? That could simplify the selected condition here (moreover if we move /new-chat and /chat to team scoped url)

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export default function TeamSelectionNavbar() {
+  const { data: teams } = useListTeamsKnowledgeFlowV1TeamsGetQuery();
+  const { pathname } = useLocation();
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles["team-navbar-container"]}>
+      <div>
+        <span className={styles.title}>{t("rework.sidebar.title")}</span>
```
florian-muller, line 16: This does not use the `siteDisplayName` (that can be retrieved with `useFrontendProperties`). We can deprecate the frontend properties but how should we handle variability in the UI ? Regardless of the way the UI get the info, I think having one central conf where all this infos are is a good thing

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
+import { useListTeamsKnowledgeFlowV1TeamsGetQuery } from "../../../../../../slices/knowledgeFlow/knowledgeFlowApiEnhancements.ts";
+import TeamSelectionItem from "@shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionItem/TeamSelectionItem.tsx";
+import styles from "./TeamSelectionNavbar.module.scss";
+import Separator from "@shared/atoms/Separator/Separator.tsx";
+import { useTranslation } from "react-i18next";
+import { useLocation } from "react-router-dom";
+
+export default function TeamSelectionNavbar() {
+  const { data: teams } = useListTeamsKnowledgeFlowV1TeamsGetQuery();
```
florian-muller, line 9: we should maybe handle teams loading (skeleton) and error ?

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+        />
+        <TeamSelectionItem
+          redirection={"/teams"}
+          teamName={t("rework.sidebar.team.allTeams")}
+          selected={pathname.startsWith(`/teams`)}
+        />
+      </div>
+      <Separator margin={"8px"} />
+      <div>
+        {teams?.map((team) => {
```
florian-muller, line 30: we should maybe handle the empty state here ? (but not sure of how we handle it)

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+          teamName={t("rework.sidebar.team.userTeam")}
+          selected={pathname.startsWith(`/agents`) || pathname.startsWith(`/knowledge`)}
+        />
+        <TeamSelectionItem
+          redirection={"/teams"}
+          teamName={t("rework.sidebar.team.allTeams")}
+          selected={pathname.startsWith(`/teams`)}
+        />
+      </div>
+      <Separator margin={"8px"} />
```
florian-muller, line 28: shouldn't the margin use a `-spacing-xs` from the theme here ?

```suggestion
      <Separator margin={"var(--spacing-xs)"} />
```

Instead of margin props being a string, it could be a enumn `"s" | "m" | "l" | ...` to enforce this

frontend/src/rework/components/shared/atoms/Separator/Separator.tsx
```diff
@@ -0,0 +1,10 @@
+import styles from "./Separator.module.scss";
+import React from "react";
+
+interface SeparatorProps {
+  margin?: string;
+}
+
+export default function Separator({ margin }: SeparatorProps) {
+  return <div className={styles["separator"]} style={{ "--separator-margin": margin } as React.CSSProperties}></div>;
```
florian-muller, line 9: Not a fan of the cast (`as`), but I understand why. Here its fine because its the only property in style but it could hide type problems. 

I was a an other solution on stackoverflow where you edit `CSSProperties` to allow property that start with `--` by putting this in the component file:
```ts
import 'react';

declare module 'react' {
    interface CSSProperties {
        // Allow any CSS variable starting with '--'
        [key: `--${string}`]: string | number
    }
}
```

Also, by curiosity, why could we not just directly set `style={{ margin }}` ? Why is it cleaner to pass by a css variable ?

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div>
+        {teams?.map((team) => {
+          return (
+            <TeamSelectionItem
+              key={team.id}
+              redirection={"/team/" + team.id}
+              teamName={team.name}
+              selected={pathname.startsWith(`/team/${team.id}`)}
+              imgUrl={"/public/images/default-team-banner.png"}
```
florian-muller, line 37: Should not start with public
```suggestion
              imgUrl={"/images/default-team-banner.png"}
```

See Vite warning for more details:
```
Files in the public directory are served at the root path.
Instead of /public/images/default-team-banner.png, use /images/default-team-banner.png.
7:08:02 PM [vite] (client) hmr update /src/rework/components/shared/layouts/MainLayout/MainLayout.tsx
7:08:05 PM [vite] (client) hmr update /src/rework/components/shared/layouts/MainLayout/MainLayout.tsx (x2)
7:08:28 PM [vite] (client) hmr update /src/rework/components/shared/layouts/MainLayout/MainLayout.tsx (x3)
```

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionItem/TeamSelectionItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+          <img
+            className={styles["team-avatar"]}
+            src={imgUrl ? imgUrl : ""}
+            alt={t("rework.sidebar.team.avatarAlt", { teamName: teamName })}
+            onLoad={() => setIsLoaded(true)}
+          />
+        </div>
+      </Link>
+      <span className={styles["team-tooltip"]}>{teamName}</span>
+    </div>
```
florian-muller, line 35: I am not sure the `isLoaded` state is needed. It meens on the first component paint, as isLoaded its false, it will show the Groups icon and then, when the image is loaded react will have to re-paint the component (its quick but it happens, I checked with the chrome performance tool).

If it just to avoid showing the image when a team has no `imgUrl`, we can just not instert the `img` when there is no `imgUrl` :

```suggestion
  const { t } = useTranslation();

  return (
    <div
      className={`${styles["team-avatar-container"]} ${selected ? styles["selected"] : ""}`}
    >
      <Link to={redirection}>
        <div className={styles["state-layer"]}>
          <span className={styles.icon}>
            <Icon category={"outlined"} type={"Groups"} filled={true} />
          </span>
          {imgUrl && <img
            className={styles["team-avatar"]}
            src={imgUrl}
            alt={t("rework.sidebar.team.avatarAlt", { teamName: teamName })}
          />}
        </div>
      </Link>
      <span className={styles["team-tooltip"]}>{teamName}</span>
    </div>
```

You also need to remove the `opacity: 0;` in `team-avatar` (and remove `loaded` class reference as it is unused).

If we want to handle the case of an image not loading, we can use the `onError` props to handle it.

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionItem/TeamSelectionItem.tsx
```diff
@@ -0,0 +1,37 @@
+import { useTranslation } from "react-i18next";
+import styles from "./TeamSelectionItem.module.scss";
+import Icon from "@shared/atoms/Icon/Icon.tsx";
+import { useState } from "react";
+import { Link } from "react-router-dom";
+
+interface TeamSelectionItemProps {
+  redirection: string;
```
florian-muller, line 8: Type could be `To` from `react-router-dom`

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <IconButton
+              size={"small"}
+              color={"primary"}
+              variant={"icon"}
+              icon={{ category: "outlined", type: "Settings", filled: true }}
+            />
+          </span>
+        </div>
+        <span className={styles["conversation-button-container"]}>
+          <ConversationButton icon={{ category: "outlined", type: "Add" }} onClick={newChatHandler}>
```
florian-muller, line 96: I think it's better to use `Link` instead of calling `navigate` in `onClick` (to support mousewheel click, right click>copy link....)

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const isUserSpace = !pathname.startsWith(`/team`);
+
+  const getTeam = () => {
+    if (teamId) {
+      return team;
+    } else {
+      return undefined;
+    }
+  };
```
florian-muller, line 72: Not sure this `getTeam` is really needed, it just handle case where a previous `team` was loaded and we go to a personal space.

We could also define a `currentTeam` that, depending on `isUserSpace` could be a hardcoded "personal team". The rest of the componant could just use the team object like if there was no special case for personal space.

```ts
const currentTeam: TeamWithPermissions = isUserSpace
    ? {
        id: "user-space",
        name: t("rework.sidebar.team.userTeam"),
        banner_image_url: "/public/images/default-team-banner.png",
      }
    : team;
```

An other solution would be to add a `key={teamId}` props when calling `TeamContentNavbar` so that it reset on every team change.

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ? `url(${getTeam().banner_image_url})`
+      : 'url("/public/images/default-team-banner.png")',
+  } as React.CSSProperties;
+
+  return (
+    <div className={styles["team-content-navbar-container"]}>
+      <div className={styles["banner-container"]} style={bannerStyle}>
+        <div className={styles["team-name-container"]}>
+          <span className={styles["team-name"]}>
+            {isLoading ? "loading" : getTeam()?.name || t("rework.sidebar.team.userTeam")}
```
florian-muller, line 84: here loading is not translated. Also, I am not sure that's that great to display "loading" ? A skeleton would be better but I am not sure you have finished them yet. 3 dots ("...") could be better ? What do you think ?

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </div>
+        <span className={styles["conversation-button-container"]}>
+          <ConversationButton icon={{ category: "outlined", type: "Add" }} onClick={newChatHandler}>
+            {t("rework.sidebar.newChat")}
+          </ConversationButton>
+        </span>
+      </div>
+      <div className={styles["navigation-container"]}>
+        <NavigationMenu items={isUserSpace ? userNavigationItems : teamsNavigationItems} />
+        <Separator margin={"16px"} />
```
florian-muller, line 103: (same, could maybe use theme spacing `--spacing-m` ?)

frontend/src/rework/components/shared/organisms/ChatList/ChatList.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const maxChat = 1;
+
+  return (
+    <div className={styles["chat-list-container"]}>
+      <div className={styles["chat-list-header"]}>
+        {t("rework.sidebar.chatList.title")}
+        <div
+          className={`${styles["chat-list-header-counter"]} ${sessions?.length >= maxChat ? styles["overflow"] : ""}`}
+          onClick={() => refetchSessions()}
```
florian-muller, line 28: Why this ? The chat counter does not have `cursor: pointer` and even if it had it, it does not seem really intuitive what it will do ? I don't think we need a refresh button for the conversation list ?

frontend/src/rework/components/shared/organisms/ChatList/ChatList.tsx
```diff
@@ -0,0 +1,43 @@
...
+        >
+          <span className={styles["overflow-icon"]}>
+            <Icon category={"outlined"} type={"Info"} />
+          </span>
+          {sessions?.length}/{maxChat}
+        </div>
+      </div>
+      <div className={styles["chat-list-items"]}>
+        {sortedSessions?.map((session) => (
+          <ChatListItem key={session.id} chat={session} refetchSessions={refetchSessions} />
```
florian-muller, line 38: I think a better react props naminb here would be to expose a `onDelete: () => void` callback from `ChatListItem`. A way for `ChatListItem` to inform its parent the delete button has been pressed, the parent can then handle this as it want. 

As here we don't want to do anything more, it would look very similar:

```tsx
<ChatListItem key={session.id} chat={session} onDelete={refetchSessions} />
```

----

An even better pattern here is, as we use RTK Query, to use their cache invalidation feature: for each route, you tell RTK Query which cache key it should use/invalidate. You can see how it was already done in `src/slices/knowledgeFlow/knowledgeFlowApiEnhancements.ts` and read more about it here: https://redux-toolkit.js.org/rtk-query/usage/automated-refetching :)

So the good pattern here would be to create an `agenticApiEnhancements.tsx` and setup all the `providesTags`/`invalidatesTags`. It would fix some problem in the whole UI !

An edge case here is that when sending a message through websocket we technicaly edit the conversation (that's when we send the first message that the agent list start being populated and we can start displaying it in the sidebar) and websocket is not handled by RTK Query. It would be nice to try to invalidate the RTK Query cache "by hand" (with code) for the current conversation each time we send a message so that the whole UI refresh (for now, mainly the list of conversations).

frontend/src/rework/components/shared/molecules/NavigationItem/NavigationItem.tsx
```diff
@@ -0,0 +1,7 @@
+interface NavigationItemProps {
+  label: string;
+}
+
+export default function NavigationItem({ label }: NavigationItemProps) {
```
florian-muller, line 5: Seem like its not used, is it a duplicate of `NavifationMenuItem` ?

frontend/src/rework/components/shared/atoms/ConversationButton/ConversationButton.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { ComponentPropsWithoutRef } from "react";
+import Icon, { IconProps } from "@shared/atoms/Icon/Icon.tsx";
+import styles from "./ConversationButton.module.scss";
+
+interface ConversationButtonProps extends ComponentPropsWithoutRef<"button"> {
+  children: React.ReactNode;
+  icon?: IconProps;
+}
```
florian-muller, line 8: Small tips, react expose a `PropsWithChildren` :)

```suggestion
interface ConversationButtonProps extends PropsWithChildren<ComponentPropsWithoutRef<"button">> {
  icon?: IconProps;
}
```
(need to be imported)

florian-muller (changes requested):
Really nice work, a full Sidebar rework and a design system creation like this is impressive, you are really good with CSS ! :)

When testing the new sidebar, I noticed a few bugs, nothing big, here they are:

- I think css to add elipsis (...) and prevent line wrap is needed on agent name in conversation list:
  Image

- When an agent name is in multiples words, the delete button can make it wrap, making the element larger:
  Image
  Image

- Really weird, when I hover some conversation it make other agent name, dates and agent icon "move" ? That really weird, I don't understand why... (good luck investigating that)
  Image

- Same as agent name, long user name and username need some css
  Image

I also noticed there is a lot of Vite warning when starting the UI (make dev), mainly this 2 :
```
Deprecation Warning [global-builtin]: Global built-in functions are deprecated and will be removed in Dart Sass 3.0.0.
Use map.get instead.

More info and automated migrator: https://sass-lang.com/d/import

...

Files in the public directory are served at the root path.
Instead of /public/images/default-team-banner.png, use /images/default-team-banner.png.
```

-----

## CSS in Fred

A small comment on how we do CSS in Fred.

Now that I see the full picture, I realize this PR moves the whole app to CSS modules, which is totally valid, just something I think we should align on as a team since it's a foundational choice. I don't know why, I just thought CSS modules would be only used in the core design system component you were adding.

Personally, I would have leaned toward Tailwind CSS as our styling solution for a new open source project like Fred. It's become the de facto standard for modern projects, has great tooling support (especially with coding agents), and is well known across the dev community. My real personal preference goes to styled-component, but the star history and npm download trends make it pretty clear that Tailwind won the CSS wars.

Image of github star history showing lots of css soltuions. A big red curve above all others show Tailwind. Second is styled-component

But as we want to define our own design system, maybe tailwind is not open enough ([I read that](https://gafemoyano.com/en/posts/from-tailwind-to-vanilla-extract/)) and our needs are more niche. I think we could explore more modern solution than sass to help us, with better type support and a better DX (in vscode it was kind of bad) ? Here are some tools/framework I heard about we could look into:
- vanilla extract: really close to css modules, but in Typescript (so types, IDE support... One less language to know)
- panda css: css in js, a bit more opinionated than vanilla extract
- unocss: light framework to create your own small tailwind (if we want to go this way)

That said, the work here is solid and I don't want to block on a stack-level discussion. I'd suggest we open a separate ticket to align on the CSS direction for the project.

In the meantime, a few tools that would pair nicely with the current CSS modules approach if we stick with it:
- `typed-css-modules` / `typed-scss-modules` -> adds type safety when referencing class names
- `clsx` -> clean conditional class name handling

PR #1410, opened 2026-03-24
Feat: Introduced new Marketplace pages, including team exploration.

### LINKED ISSUE #1382: MAJ UI Page marketplace
- Ajout du panneau de navigation (Equipes, Lumis (coming soon)) #front
- Rework du contenu de la page avec le nouveau Design System #front

Felix-Page:
- Introduced new Marketplace pages, including team exploration.
- Added `TeamCard`, `AvatarGroup`, and `Tooltip` reusable


fred-core/fred_core/tests/integration/test_rebac.py
```diff
@@ -459,12 +459,12 @@ async def test_team_hierarchy_and_permissions(
     # Manager
 
     # Test manager can update members
```
florian-muller:
```suggestion
    # Test manager can not update members
```

frontend/src/common/router.tsx
```diff
@@ -89,6 +90,10 @@ export const routes: RouteObject[] = [
         path: "team/:teamId/*",
         element: <TeamDetailsPage />,
       },
+      {
```
florian-muller, line 88: Old "/teams" route and component (`TeamsPage`) were not removed, should'n we remove them ?

frontend/src/pages/TeamDetailsPage.tsx
```diff
@@ -71,36 +70,6 @@ export function TeamDetailsPage() {
...
-              maxWidth: "90ch",
-            }}
-          >
-            {team?.description}
-          </Typography>
-        </Box>
-      </Box>
-
       {/* Tabs */}
       <NavigationTabs
```
florian-muller, line 66: As now there is no Tabs to naviguate between the different team pages (Agents, Resources, Apps...) I don't think `NavigationTabs` has still a purpose here, we could put the routes in the main router (router.tsx).

It also allows access to old UI at /members, /apps... (that sould be deleted too ?)

I think we should even remove `TeamDetailsPage` entirely as there is no need for a common header before all pages

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,20 @@
+import Button from "@components/shared/atoms/Button/Button";
+import styles from "./TeamAgentsPage.module.scss";
+import { useTranslation } from "react-i18next";
+
+export default function TeamAgentsPage() {
```
florian-muller: This is not used ?

frontend/src/rework/components/pages/marketplace/MarketplaceTeams/MarketplaceTeams.tsx
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+      <div className={styles["marketplace-teams-content"]}>
+        <div className={styles["marketplace-teams-list-subtitle"]}>{t("rework.marketplace.teams.yourTeams")}</div>
+        <div className={styles["marketplace-teams-list"]}>
+          {yourTeams && yourTeams.map((team) => <TeamCard key={team.id} team={team} withDescription={false} />)}
+        </div>
+        <div className={styles["marketplace-teams-list-subtitle"]}>{t("rework.marketplace.teams.otherTeams")}</div>
+        <div className={styles["marketplace-teams-list"]}>
+          {otherTeams && otherTeams.map((team) => <TeamCard key={team.id} team={team} withDescription={true} />)}
+        </div>
```
florian-muller, line 26: Empty state is not handled and as for now the public team feature is not working, I think we sould handle it. Maybe hide the whole section if `otherTeams` is empty ?

frontend/src/rework/components/shared/organisms/TeamCard/TeamCard.module.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.team-card-description {
+  display: -webkit-box;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  color: var(--on-surface-retreat);
+
+  @include typography("body", "medium");
+}
```
florian-muller, line 86:
```suggestion
.team-card-description {
  display: -webkit-box;
  -webkit-line-clamp: 3;
  -webkit-box-orient: vertical;
  text-overflow: ellipsis;
  overflow: hidden;
  color: var(--on-surface-retreat);

  @include typography("body", "medium");
}
```

We could add `text-overflow: ellipsis;` to show `...` for description with one long big line with no space, like:
Screenshot from 2026-03-25 19-41-18

frontend/src/rework/components/shared/atoms/UserAvatar/UserAvatar.tsx
```diff
@@ -1,15 +1,20 @@
...
-export default function UserAvatar({ name, size }: UserAvatarProps) {
+export default function UserAvatar({ name, size, tooltip = false, ...props }: UserAvatarProps) {
   return (
-    <div className={styles["user-avatar"]} data-size={size}>
+    <div className={styles["user-avatar"]} data-size={size} {...props}>
       {getInitials(name)}
+      {tooltip && (
+        <span className={styles["user-avatar-tooltip"]}>
+        </span>
+      )}
```
florian-muller: Maybe I don't understand but I am under the impression this does nothing ? Tooltip seems to be handled by the `Tooltip` component wrapping the `UserAvatar`.

If that is the case, we should remove the `tooltip` props from `UserAvatar`, or put the wrapping `Tooltip` in it and optionaly wrap it depending on `tooltip` props

frontend/src/rework/components/shared/molecules/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <div className={styles["user-avatar-container"]}>
+      {avatars.map((avatar, index) => (
+        <Tooltip key={index} text={avatar.name}>
+            <UserAvatar tooltip={tooltip} {...avatar} />
+        </Tooltip>
+      ))}
+    </div>
+  );
+}
```
florian-muller: (As sayd in the last comment, it seems like `UserAvatar`'s `tooltip` props does nothing, so changing `AvatarGroup`'s `tooltip` here does not hide the tootltip)

frontend/src/rework/components/shared/organisms/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <div className={styles["team-member-count"]}>
+            <span className={styles["team-member-count-icon"]}>
+              <Icon category={"outlined"} type={"groups"} />
+            </span>
+            {t("rework.teamCard.memberCount", { count: team.member_count })}
+          </div>
+        </div>
+        {withDescription && <div className={styles["team-card-description"]}>{team.description}</div>}
+        <div className={styles["team-card-footer"]}>
+          <AvatarGroup tooltip={true} avatars={team.owners.map((o) => ({ name: o.first_name + " " + o.last_name, size: "small" }))} />
```
florian-muller:
```suggestion
          <AvatarGroup tooltip avatars={team.owners.map((o) => ({ name: o.first_name + " " + o.last_name, size: "small" }))} />
```
Just a preference, you can just write `tooltip` instead of `tooltip={true}`. I was under the impression that the shorter version was more idiomatic but that's not that important

frontend/stats.html
florian-muller, line 1: Should we commit this file ?

frontend/src/rework/components/shared/atoms/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,42 @@
+import { cloneElement, CSSProperties, HTMLAttributes, ReactElement, useId, useRef } from "react";
```
florian-muller, line 1: Not a fan of `cloneElement`, not very type safe. It means the children should be able to handle props like `style`, `onMouseEnter`... and if not, it fails silently and there is no type error.

At the type level this should not work on `UserAvatar`, it doesn not have these props. It only works because of there is a `...props` that is deconstructed in the root div. But having a `...props` here although all the props were already deconsctructed is weird ? `props` type is `(parameter) props: {}` (from IDE), it looks like it does nothing, but silently it make the tooltip work.

----
### Soltions

But, I have looked a bit at popular tooltip implementations (like Radix UI, MUI...) and it looks like they all do something like that. 

#### Wrapping

By default, RadixUI add a wrapping `button` on the trigger of the tooltip. It allows it to safely add the props it needs without changing the children. If you don't want the additional button in the DOM, you can pass `asChild` and it will do a cloneElement on the children, like you did, with the same type problem (but as it is an "opt-in" feature, its maybe a bit more explicit that your trigger should support some props ?). 

#### Explicit typing

A more radical changes would be to change `children` type to a function taking the required props and returning a `ReactElement`, like:

```tsx
export interface TooltipTriggerProps {
  style: CSSProperties;
  onMouseEnter: () => void;
  onMouseLeave: () => void;
  onFocus: () => void;
  onBlur: () => void;
}

interface TooltipProps {
  text: string;
  children: (props: TooltipTriggerProps) => ReactElement;
}
```

When using tooltip, if you pass an other react element direclty you will get an error:
image

You have to pass a real function, and can get type error if the children does not support it:
image

But if you try to pass trigger props with the spread opperator (`...`), you don't get type error, so I am not sure its worth it:
image

Of course, you have to adapt a bit Tooltip. and you have no `cloneElement`:
```tsx
  const triggerProps: TooltipTriggerProps = {
    style: { anchorName } as CSSProperties,
    onMouseEnter: () => popoverRef.current?.showPopover(),
    onMouseLeave: () => popoverRef.current?.hidePopover(),
    onFocus: () => popoverRef.current?.showPopover(),
    onBlur: () => popoverRef.current?.hidePopover(),
  };

  return (
    <>
      {children(triggerProps)}

      // ...
    <>
  )
```

#### Child tips

One slight potential improvement at the children level (like `UserAvatar` here) would be to make the props extends HTMLDivElement so that when writing `...props`, props has the right types a div could recevive (and you will get a type error if you pass it to the wrong element):

```typescript
interface UserAvatarProps extends HTMLAttributes<HTMLDivElement> {
  name: string;
  size: "x-small" | "small" | "medium" | "large";
  tooltip?: boolean;
}

```

To make sure all our components are compatible with the Tooltip, we could make this a norm and add it to all of our components (the atoms at least) ?

frontend/src/rework/components/shared/organisms/Sidebar/Sidebar.tsx
```diff
@@ -3,25 +3,28 @@ import TeamContentNavbar from "@shared/organisms/Sidebar/TeamContentNavbar/TeamC
...
+        <UserProfile />
+      </div>
     </div>
   );
 }
+
+enum SidebarMode {
+  TEAM,
+  MARKETPLACE,
+}
```
florian-muller: I always heard that Enum in Typescript were bad (generate runtime code, map to numbers by default creating weird behaviour), maybe we could use a string union type instead (zero runtime cost + full type safety)

```suggestion
type SidebarMode = "team" | "marketplace";
```

frontend/src/rework/components/shared/organisms/Sidebar/MarketplaceNavbar/MarketplaceNavbar.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export default function MarketplaceNavbar() {
+  const { t } = useTranslation();
+
+  const navigationItems: NavigationMenuItemProps[] = [
+    {
+      type: "link",
+      label: t("rework.sidebar.marketplace.menu.teams"),
+      icon: { category: "outlined", type: "groups", filled: true },
+      selected: false,
```
florian-muller: Not a fan of this `selected: false`, when reading code, it give the impression that this item will never be selected. If we never need it for link type MenuItem, we could remove it from the type entrirely:

```tsx
export interface NavigationMenuItemBase {
  label: string;
  icon: IconProps;
}

export interface NavigationLinkProps extends NavigationMenuItemBase {
  type: "link";
  linkProps: LinkProps;
}

export interface NavigationActionProps extends NavigationMenuItemBase {
  type: "button";
  selected: boolean; // selected moved here
  onClick: () => void;
}
```

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -23,41 +23,17 @@ export default function TeamContentNavbar() {
...
   const canOpenTeamSettings = selectedTeam?.permissions?.includes("can_administer_owners") || false;
 
-  const teamsNavigationItems: NavigationMenuItemProps[] = [
+  const navigationItems: NavigationMenuItemProps[] = [
     {
       type: "link",
-      label: t("rework.sidebar.menu.agents"),
+      label: t("rework.sidebar.team.menu.agents"),
       icon: { category: "outlined", type: "Person" },
       selected: pathname.startsWith(`/team/${teamId}/agents`),
```
florian-muller: (selected should be removed here too, as they are not used)

frontend/src/rework/components/shared/molecules/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,20 @@
+import styles from "./AvatarGroup.module.scss";
+import UserAvatar, { UserAvatarProps } from "@shared/atoms/UserAvatar/UserAvatar.tsx";
+import {Tooltip} from "@shared/atoms/Tooltip/Tooltip.tsx";
+
+interface AvatarGroupProps {
+  avatars: UserAvatarProps[];
+  tooltip?: boolean;
+}
+
+export default function AvatarGroup({ avatars, tooltip = false }: AvatarGroupProps) {
```
florian-muller: We may need to add a way to handle big number of avatars:

Image

Maybe like we had in MUI, with "+4" avatar if we have more than a given treshold ?

frontend/src/rework/components/shared/organisms/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+        className={styles["team-banner"]}
+        src={team.banner_image_url ? `url(${team.banner_image_url})` : "/images/default-team-banner.png"}
+        alt=""
+        aria-hidden="true"
+      ></img>
+      <img className={styles["team-avatar"]} src="/images/defaultTeamAvatar.png" alt="" aria-hidden="true"></img>
+      <div className={styles["team-card-details"]}>
+        <div className={styles["team-card-detail-name"]}>
+          <div className={styles["team-information"]}>
+            <div className={styles["team-name"]}>{team.name}</div>
```
florian-muller, line 27: Not that important as for now we are creating team ourself, but long team name are not handled: 

Image

florian-muller (commented):
Clean work as always ! :)

frontend/src/pages/TeamDetailsPage.tsx
florian-muller: You @Felix-Page think we should keep TeamDetailsPage ?

frontend/src/rework/components/pages/marketplace/MarketplaceTeams/MarketplaceTeams.tsx
florian-muller: @Felix-Page is it not a good idea to hanlde this ?

frontend/src/rework/components/shared/atoms/Tooltip/Tooltip.tsx
florian-muller: @Felix-Page any preferences/thoughts on this ?

frontend/src/rework/components/shared/molecules/AvatarGroup/AvatarGroup.tsx
```diff
@@ -1,18 +1,18 @@
...
   avatars: UserAvatarProps[];
-  tooltip?: boolean;
 }
 
-export default function AvatarGroup({ avatars, tooltip = false }: AvatarGroupProps) {
+export default function AvatarGroup({ avatars }: AvatarGroupProps) {
   return (
     <div className={styles["user-avatar-container"]}>
-      {avatars.map((avatar, index) => (
+      {avatars.length > 4 && <UserAvatar name={`+ ${(avatars.length - 4).toString()}`} size={"small"} />}
```
florian-muller, line 12: I don't think we should force size small on `UserAvatar`:
Image

Shouldn't we pass `{...avatar}` ?

florian-muller (dismissed):

florian-muller (approved):

PR #1417, opened 2026-03-25
Feat: add patch for enabled stateless connection with MCP protocol

### LINKED ISSUE #1408: Knowledge Flow pod cannot be horizontally scaled due to stateful MCP session handling
The Knowledge Flow pod hosts MCP servers using the fastapi_mcp library (tadata-org/fastapi_mcp). Because the MCP protocol is stateful (session-based), it is currently impossible to run multiple replicas of this pod — all requests from a given client must reach the same instance, preventing horizontal scaling.

The fastapi_mcp library hardcodes stateless=False in its HTTP transport layer (fastapi_mcp/transport/http.py), meaning sessions are always maintained. There is no configuration option exposed to override this behavior.

By contrast, the upstream fastmcp library (PrefectHQ/fastmcp) does support a stateless_http=True mode that enables horizontal scaling — but only if sampling is not used (see fastmcp#678: sampling hangs when stateless_http=True).

We do not appear to use sampling at this time, which means stateless_http=True would be viable if we were using fastmcp directly.
However, we use fastapi_mcp, which does not expose this option.

sebastienehlingfred:
## Summary

Describe what changed and why.

## Mandatory Checklist

- [ ] I followed [`docs/DEVELOPER_CONTRACT.md`](../docs/DEVELOPER_CONTRACT.md).

## Validation Notes

## Risk / Rollback

### DIFF
--- a/knowledge-flow-backend/knowledge_flow_backend/compat/fastapi_mcp_patch.py
+++ b/knowledge-flow-backend/knowledge_flow_backend/compat/fastapi_mcp_patch.py
@@ -1,15 +1,33 @@
 """
-Monkey patch for fastapi_mcp's schema resolver to avoid infinite recursion.
+Monkey patches for fastapi_mcp.
 
-See https://github.com/tadata-org/fastapi_mcp/pull/156 for upstream fix.
-TODO: Remove this file once the project depends on a release that includes that PR.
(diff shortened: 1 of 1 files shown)


florian-muller (approved):
Well done, seems to fix the problem we had when running multiples KF replicas

PR #1418, opened 2026-03-25
1363 rework UI liste des lumis

### LINKED ISSUE #1363: Rework UI : Liste des Lumis
Mettre à jour l'UI de la liste des Lumis avec le nouveau Design System #front

Felix-Page:


dimitri-tombroff (dismissed):

frontend/src/locales/fr/translation.json
```diff
@@ -1402,6 +1404,12 @@
         }
       }
     },
+    "teams":{
+      "agents":{
+        "title": "Sollicitez vos Agents",
+        "create": "Créer un Agent"
+      }
+    },
```
florian-muller, line 1412: It seems to me that none of the translation key addition/deletion in this file have been added to the English version ?

image

----

I personaly use the "i18n Ally" vscode extension to help me work with i18n. It shows me warning in my IDE if I forgot a translation and add quick action to edit them (without having to go in the json by hand). I checked and there is a plugin for Jetbrains :)
https://plugins.jetbrains.com/plugin/17212-i18n-ally

image

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -15,12 +14,11 @@ import TeamSettingsPage from "@components/pages/TeamSettingsPage/TeamSettingsPag
 export default function TeamContentNavbar() {
   const [isTeamSettingsOpen, setIsTeamSettingsOpen] = useState(false);
   const { t } = useTranslation();
-  const navigate = useNavigate();
   const { teamId } = useParams<{ teamId: string }>();
 
-  const { data: team } = useGetTeamQuery({ teamId: teamId !== "user" ? teamId : "" }, { skip: !teamId });
+  const { data: team } = useGetTeamQuery({ teamId: teamId }, { skip: !teamId || teamId === 'personal'});
```
florian-muller, line 19: Instead of hardcoding `"personal"` here, why not using `userDetails?.personalTeamId` like in `TeamSelectionNavbar` ?

Something like that:
```ts
const { teamId } = useParams<{ teamId: string }>();
const { data: userDetails } = useGetUserDetailsControlPlaneV1UserGetQuery();

const isPersonalTeam = teamId === userDetails?.personalTeamId 
```

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -17,9 +17,9 @@ export default function TeamSelectionNavbar() {
       <div>
         <span className={styles.title}>{t("rework.sidebar.title")}</span>
         <TeamSelectionItem
-          redirection={"/team/" + userDetails?.personalTeamId}
+          redirection={`/team/${userDetails?.personalTeamId}/agents`}
           teamName={t("rework.sidebar.team.userTeam")}
-          selected={pathname.startsWith(`/team/user`)}
+          selected={pathname.startsWith(`/team/personal`)}
```
florian-muller, line 22: Should it use `personalTeamId` ? Like:

```suggestion
          selected={pathname.startsWith(`/team/${userDetails?.personalTeamId}`)}
```

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { useGetTeamQuery } from "../../../../slices/controlPlane/controlPlaneApi.ts";
+import { useMemo, useState } from "react";
+import { useAgentUpdater } from "../../../../hooks/useAgentUpdater.ts";
+import { AnyAgent } from "../../../../common/agent.ts";
+import { AgentCreateEditDrawer } from "../../../../components/agentHub/AgentCreateEditDrawer.tsx";
+
+export default function TeamAgentsPage() {
+  const { t } = useTranslation();
+  const { teamId } = useParams();
+  const { data: agents, refetch } = useListAgentsAgenticV1AgentsGetQuery({ ownerFilter: "team", teamId });
```
florian-muller, line 16: As personal team does not exist yet, this does not work on personal space. To retrieve them, in the current api, we need to pass `ownerFilter: "personal"`.

I see 2 solutions:
- We add a `isPersonal` here (like in other components, by comparing with `personalTeamId`) and use it to choose the right param
    ```ts
    const { data: agents, refetch } = useListAgentsAgenticV1AgentsGetQuery({ ownerFilter: isPersonal ? "personal" : "team", teamId });
    ```

- We patch all team endpoint to hande the `team` id in the same way it handled the `ownerFilter: "personal"` -> the special "personal" team case is only hamdled at backend side and the front code will not have to change when we switch to real team

What do you think ?

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { data: team } = useGetTeamQuery({ teamId: teamId !== "user" ? teamId : "" }, { skip: !teamId });
+  const { updateEnabled } = useAgentUpdater();
+  const [selected, setSelected] = useState<AnyAgent | null>(null);
+  const [editOpen, setEditOpen] = useState(false);
+
+  const canUpdateAgents = useMemo(() => {
+    return team?.permissions?.includes("can_update_agents");
+  }, [team]);
```
florian-muller, line 24: Same here, there is no team with id `personal` so `useGetTeamQuery` return 404 and `team` here stay undefined and you never get the right to update agent.

Same kind of solutions I think: we could return a fake team with id `personal` to write the frontend cleanly or handle it quickly here with a `isPersonal`

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const handleToggleEnabled = async (agent: AnyAgent) => {
+    const isEnabled = agent.enabled !== false;
```
florian-muller: We could use nullish coalescing operator here:

```suggestion
    const isEnabled = agent.enabled ?? true;
```

I personaly find it a bit more clear to read.

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+            </Button>
+          )}
+        </div>
+        <div className={styles.agentList}>
+          {agents?.map((agent) => (
+            <>
+              <AgentCard
+                key={agent.id}
+                agent={agent}
+                readOnly={canUpdateAgents}
```
florian-muller: This is not possible with the current backend, but I would just like to note that in the future, I think it will be better to test if there is an "update" in `agent.permission` instead of relying on `canUpdateAgents` (which, now that I am thinking of it, should just be `can_create_agent` and all other checks should be done on the agents permission directly).

This would allow to not have to rewrite this logic if one day users can see agent they can update and agent they can't update in the same team (like, for example, if they added a readonly agent from a marketplace... wink wink).

But I understand that adding a `permission` field to the agent api is outiside of this PR scope :)


```suggestion
                # todo: in future, rely on direct `update` and `delete` permissions from agent (when they are returned by backend)
                readOnly={canUpdateAgents}
```

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+    setEditOpen(true);
+  };
+
+    const handleEdit = (agent: AnyAgent) => {
+      setSelected(agent);
+      setEditOpen(true);
+    };
+
+    return (
+      <div className={styles.teamAgentContainer}>
```
florian-muller: There is no padding on this container:

Image

Same on the bottom:

Image

Also there is no max width, but that's might not be a problem (@Maxime-DARAGON  ?)

Image

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+              size={"medium"}
+              icon={{ category: "outlined", type: "add" }}
+              onClick={handleOpenCreateAgent}
+            >
+              {t("rework.teams.agents.create")}
+            </Button>
+          )}
+        </div>
+        <div className={styles.agentList}>
+          {agents?.map((agent) => (
```
florian-muller: When I enabled / disabled some agents, they were sometimes shifitng places (and sometimes not, did not understood why). Maybe we could sort alphabetically this `agents` list ?

```ts
const sortedAgents = [...(agents ?? [])].sort((a, b) => a.name.localeCompare(b.name));
```

florian-muller (changes requested):
Nice work, lovely effect on the hover of the agent card !

This is pretty good, just a few tiny bug to fix before the merge, mains one being personal agent that seem broken and a few missing translation.

I also encountered the following bug when testing but had not time to investigate further, so not sure this is coming from this PR:

- The fields in the form to create/edit agent are super laggy, like it takes 1s after a key press to update. Seems like there is a bad state or something wired here

we can check this together if you want

frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
Felix-Page: I do this in the next story, with the team's conversation scoping

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
Felix-Page: I do this in the next story, with the team's conversation scoping

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
Felix-Page: I do this in the next story, with the team's conversation scoping

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
Felix-Page: I do this in the next story, with the team's conversation scoping

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
Felix-Page: It doesn't have the same meaning. I consider that the fact an agent is deactivated should be sent to me; the value shouldn't be undefined.

florian-muller (dismissed):

florian-muller (approved):

PR #1462, opened 2026-04-02
1441 Scope conversations by teams and fix personal team workflow

### LINKED ISSUE #1441: Scopper les conversations par equipe
- Scopper les conversations par équipe
- Lancement d'une conversation dans le contexte d'une equipe
- Mise en place de la route frontend du chat dans le contexte de l'équipe

Felix-Page:
Implement team-based session filtering, fix personal team workflow and remove legacy frontend components


control-plane-backend/control_plane_backend/teams_controller.py
```diff
@@ -104,6 +104,15 @@ async def get_team(
...
 ) -> TeamWithPermissions:
+    if team_id == "personal":
+        return TeamWithPermissions(
+            id="personal",
+            name="Equipe personnelle",
+            member_count=1,
+            is_private=True,
+            owners=[],
+            permissions=["can_read", "can_update_resources", "can_update_agents"]
+        )
```
florian-muller: Why put this here and not in the service ? What if one day we call `get_team_by_id_from_service` in an other place ? I think controller should only be used for HTTP REST API logic

control-plane-backend/control_plane_backend/users_controller.py
```diff
@@ -85,12 +91,23 @@ async def delete_user(
...
+) -> UserDetails:
+    return UserDetails (
+        personalTeam=TeamWithPermissions(
+            id="personal",
+            name="Equipe personnelle",
+            member_count=1,
+            is_private=True,
+            owners=[],
+            permissions=["can_read", "can_update_resources", "can_update_agents"]
+        )
```
florian-muller, line 117: To be abit more DRY, maybe we could define this personal team one time and import it to use it everywhere (to avoid future drift in the personal mock team)

frontend/src/rework/components/shared/organisms/ChatList/ChatListItem/ChatListItem.tsx
```diff
@@ -10,19 +10,20 @@ import { useTranslation } from "react-i18next";
 import React from "react";
 
 interface ChatListItemProps {
+  teamId: string;
   chat: SessionWithFiles;
```
florian-muller: isn't team id already present in `chat.tema_id` ?

frontend/src/common/router.tsx
```diff
@@ -55,22 +54,18 @@ export const routes: RouteObject[] = [
     children: [
       {
         index: true,
-        element: <Navigate to="/new-chat" replace />,
-      },
-      {
-        path: "/new-chat",
-        element: <NewChatAgentSelection />,
+        element: <Navigate to="/team/personal/agents" replace />,
```
florian-muller, line 57: For now it works but if one day personal team have real id (and not personal), it will not. To be clean, we should get the personal team with the user route (a bit much for a redirect but...)

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionNavbar.tsx
```diff
@@ -17,7 +17,7 @@ export default function TeamSelectionNavbar() {
       <div>
         <span className={styles.title}>{t("rework.sidebar.title")}</span>
         <TeamSelectionItem
-          redirection={`/team/${userDetails?.personalTeamId}/agents`}
+          redirection={`/team/${userDetails?.personalTeam.id}/agents`}
           teamName={t("rework.sidebar.team.userTeam")}
           selected={pathname.startsWith(`/team/personal`)}
```
florian-muller: frontend should not use `/personal`
```suggestion
          selected={pathname.startsWith(`/team/${userDetails?.personalTeam.id}`)}
```

agentic-backend/agentic_backend/core/chatbot/session_orchestrator.py
```diff
@@ -1061,14 +1061,15 @@ async def resume_interrupted_exchange(
...
         user: KeycloakUser,
+        team_id: str,
     ) -> List[SessionWithFiles]:
         """
-        Get all sessions for a user, enriched with the list of uploaded files.
+        Get all sessions for a user filtered by a team, enriched with the list of uploaded files.
         This method is only used by the UI to list sessions. It is not part of the
         chat exchange flow.
         """
         async with phase_timer(self.kpi, "session_list"):
```
florian-muller, line 1083: I think we should check here that the user is really still a member of the team. Without it, it could allow a user who was deleted from team to access its old conversation.

As Team membership are handled by Keycloak for now, we can just do a check like this:
```python
        if team_id not in user.groups:
            raise AuthorizationError(
                user.uid,
                Action.READ,
                Resource.SESSIONS,
                f"Not authorized to read sessions for team {team_id}",
            )
```

I would have prefered to call a function from the team service to do it (to have all team logic in one place, I fear that when removing Keycloak from team we forgot this small piece of code lost here) but as it is in an other backend...

florian-muller (changes requested):
Nice work, a real pleasure to have the personal space back and the conversation scopped

A few thing before we merge:

- There is multiple types error in the python part (not catched by the CI as the control plane CI was not done). You can see them by running `make type-check`, but an even better way would be to have the type checker configured in you IDE directly (so you see the error as you are writing them). For information, we use pyright as type checker, it seems like there is a plugin for jetbrain: https://plugins.jetbrains.com/plugin/24145-pyright but not sure if that's the right way to set it up

- As already discusses, the /preference route overriding the teamId (and allowing unsecrue access)

frontend/src/common/router.tsx
Felix-Page: I think that when the real personal teams arrive, we'll stick to this path and manage things pretty much the same way they're currently set up. For now, let's keep this solution and we'll discuss it again during the final implementation.

control-plane-backend/control_plane_backend/users_controller.py
```diff
@@ -86,11 +91,28 @@ async def delete_user(
...
-) -> dict[str, str]:
-    return {"personalTeamId": "personal"}
+) -> UserDetails:
+    return UserDetails(
+        personalTeam=TeamWithPermissions(
+            id=TeamId("personal"),
+            name="Equipe personnelle",
+            member_count=1,
+            is_private=True,
+            owners=[],
```
florian-muller: shouldn't we put the current user as owner of its own team ?

frontend/src/pages/TeamDetailsPage.tsx
```diff
@@ -6,12 +6,18 @@ import { TeamAppsPage } from "../components/teamDetails/TeamAppsPage";
...
   const { t } = useTranslation();
 
   const { teamId } = useParams<{ teamId: string }>();
-  const { data: team, isLoading } = useGetTeamQuery({ teamId: teamId !== "user" ? teamId : "" }, { skip: !teamId });
+  const { data: userDetails } = useGetUserDetailsControlPlaneV1UserGetQuery();
+  const { data: fetchedTeam, isLoading } = useGetTeamQuery(
+    { teamId: teamId },
+    { skip: !teamId || teamId === userDetails?.personalTeam.id },
+  );
+  const team = fetchedTeam ?? userDetails?.personalTeam;
```
florian-muller: when real team is loading (so `fetchedTeam` is undefined), `userDetails?.personalTeam` will be used

frontend/src/rework/components/pages/TeamAgentsPage/TeamAgentsPage.tsx
```diff
@@ -2,18 +2,21 @@ import Button from "@components/shared/atoms/Button/Button";
...
 export default function TeamAgentsPage() {
   const { t } = useTranslation();
   const { teamId } = useParams();
-  const { data: agents, refetch } = useListAgentsAgenticV1AgentsGetQuery({ ownerFilter: "team", teamId });
+  const { data: userDetails } = useGetUserDetailsControlPlaneV1UserGetQuery();
+  const ownerFilter = teamId === userDetails?.personalTeam.id ? "personal" : "team";
+  const { data: agents, refetch } = useListAgentsAgenticV1AgentsGetQuery({ ownerFilter: ownerFilter, teamId });
   const { data: team } = useGetTeamQuery({ teamId: teamId !== "user" ? teamId : "" }, { skip: !teamId });
   const { updateEnabled } = useAgentUpdater();
   const [selected, setSelected] = useState<AnyAgent | null>(null);
```
florian-muller, line 28: Not on this line but on useMemo a few line down:

```ts
 const canUpdateAgents = useMemo(() => {
    return team?.permissions?.includes("can_update_agents");
  }, [team]);
```

`canUpdateAgents` is only computed on `team` and only recomputed on `team` change so when you arrive on the personal agent space, you don't have the button (as no team is loaded at this time). When you go to a team (load one, `team` is not undefine) and come back, you have the button to create the team

florian-muller (changes requested):
Perfect, just a small blocking bug where the button to create agent does not appear when you load the page on your personal team

florian-muller (approved):

florian-muller (changes requested):
bug: chat history not loading in editor ? need to investigate

agentic-backend/agentic_backend/core/chatbot/session_orchestrator.py
```diff
@@ -1061,14 +1061,22 @@ async def resume_interrupted_exchange(
...
-            sessions = await self.session_store.get_for_user(user.uid)
+            if team_id not in user.groups:
```
florian-muller: In fact this does not work, team_id is a real id (like `'1151bdfa-d48e-4cca-96b1-a0ccfc798c30'`) and `user.groups` contains name of Keycloak groups (`'/fredlab', '/northbridge'`). This make fetching sessions impossible

florian-muller (dismissed):

florian-muller (approved):

PR #1482, opened 2026-04-07
Setup branding in new UI

Felix-Page:


frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -28,22 +31,22 @@ export default function TeamContentNavbar() {
   const navigationItems: NavigationMenuItemProps[] = [
     {
       type: "link",
-      label: t("rework.sidebar.team.menu.agents"),
-      icon: { category: "outlined", type: "Person" },
+      label: agentsNicknamePlural.toLowerCase().replace(/\b\w/g, (char) => char.toUpperCase()),
+      icon: { category: "outlined", type: agentIconName as IconType },
```
florian-muller: not a fan of the cast. If we accept any string, maybe `type` should be as tring here. If we access a small subset of string, agentIconName should have a better type ?

florian-muller (approved):
LGTM

PR #1488, opened 2026-04-08
Fix banner upload + update design

Felix-Page:


frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx
```diff
@@ -32,21 +33,22 @@ export default function TeamContentNavbar() {
...
     {
       type: "link",
       label: t("rework.sidebar.team.menu.resources"),
-      icon: { category: "outlined", type: "folder" },
+      icon: { category: "outlined", type: "folder", filled: true },
       linkProps: { to: `/team/${teamId}/resources` },
     },
   ];
-
+console.log(teamId, userDetails?.personalTeam.id);
```
florian-muller:
```suggestion

```
should be removed ?

frontend/src/rework/components/shared/organisms/Sidebar/TeamSelectionNavbar/TeamSelectionItem/TeamSelectionItem.tsx
```diff
@@ -20,25 +20,37 @@ export default function TeamSelectionItem({
   icon = { category: "outlined", type: "groups", filled: true },
 }: TeamSelectionItemProps) {
   const { t } = useTranslation();
-  const [isLoaded, setIsLoaded] = useState(false);
+  const tooltipId = useMemo(() => `tooltip-${teamName}`.replace(" ", "_"), [teamName]);
```
florian-muller: I thibk this can cause a problem if 2 teams have the same name (which is not possible for now because Keycloak does not allow it, but could happen in the future when we move out of Keycloak)

florian-muller (changes requested):
I think banner on TeamCard are missing `object-fit: cover;` 
Image

-----

Not sure why, my really long team name tooltip is now centered: 
Image

That's ok as we name the team ourself, just wanted to show you

florian-muller (approved):

PR #1608, opened 2026-05-07
1607 fix: race condition in frontend when loading frontend_settings from the agentic backend

### LINKED ISSUE #1607: fix Cannot read properties of undefined (reading 'toLowerCase')
race condition when loading the frontend settings causing objects to be null when accessing toLowerCase property

Title: Frontend crashes on initial render when frontend properties are unavailable or truncated

Description

In production, the frontend can crash with:

```text
Unexpected Application Error!
can't access property "toLowerCase", agentsNicknamePlural is undefined
```

Observed behavior:
- the crash happens on the initial page load
- refreshing the page often makes the issue disappear
- this was observed on the frontend pod running on Kubernetes

Root cause

This is a first-render race around frontend configuration properties.

Two contributing factors were identified:

1. `useFrontendProperties()` can expose an empty or incomplete object during render
2. `normalizeProps()` in `frontend/src/common/config.tsx` did not preserve all fields from `frontend_settings.properties`

Because of that, components such as:

- `frontend/src/rework/components/shared/organisms/Sidebar/TeamContentNavbar/TeamContentNavbar.tsx`

can execute code like:

```ts
agentsNicknamePlural.toLowerCase()
```

while `agentsNicknamePlural` is still `undefined`.

Why this is intermittent

A page refresh often fixes the issue because the configuration fetch completes before the next render path reaches the sidebar. The bug is therefore timing-dependent rather than data-dependent.

Impact

- user-facing crash on page load
- intermittent and hard-to-reproduce behavior
- production instability on cold loads / first renders

Suggested fix

Use the already bootstrapped frontend config as the single source of truth during render, instead of relying on a second async fetch in `useFrontendProperties()`.

Also ensure that `normalizeProps()` preserves all supported frontend properties, including at least:

- `agentsNicknamePlural`
- `agentsNicknameSingular`
- `agentIconName`
- `siteTitle`
- `siteSubtitle`
- `defaultTeamBannerFile`
- `defaultPersonalBannerFile`
- `defaultTeamAvatarFile`
- `defaultPersonalAvatarFile`
- `gcuVersion`
- `uploadWarning`

Acceptance criteria

- no frontend crash on initial load
- no `toLowerCase()` error while rendering sidebar/navigation
- stable behavior without requiring a manual page refresh
- frontend build passes

Reproduction hints

- deploy the frontend in a production-like environment
- open the app in a fresh session / cold load
- observe the intermittent crash before any refresh
- refresh the page and observe that the app often loads normally

simcariou:
## Summary

Describe what changed and why.

## Mandatory Checklist

- [ ] I followed [`docs/platform/DEVELOPER_CONTRACT.md`](../docs/platform/DEVELOPER_CONTRACT.md).

## Validation Notes

## Risk / Rollback

### DIFF
--- a/frontend/src/hooks/useFrontendProperties.ts
+++ b/frontend/src/hooks/useFrontendProperties.ts
@@ -12,31 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { useMemo } from "react";
-import { useGetFrontendConfigAgenticV1ConfigFrontendSettingsGetQuery } from "../slices/agentic/agenticOpenApi";
+import { getConfig } from "../common/config";
 import type { Properties } from "../slices/agentic/agenticOpenApi";
 
 /**
- * Custom hook to access frontend properties from the configuration.
+ * Custom hook to access frontend properties from the bootstrapped frontend configuration.
  *
- * This hook wraps the frontend config query and exposes only the properties object.
- * The query is configured with aggressive caching since frontend config rarely changes
- * (only on helm chart redeployment).
+ * Why this exists:
+ * - Frontend properties are already loaded by `loadConfig()` before React mounts.
+ * - Reading the bootstrapped config avoids a second async fetch and removes first-render races in navigation components.
+ *
+ * How to use it:
+ * - Call the hook inside a React component and destructure the properties you need.
+ * - The hook returns the same backend-provided properties stored in the shared frontend config.
+ *
+ * Example:
+ * - `const { agentsNicknamePlural, siteDisplayName } = useFrontendProperties();`
  *
  * @returns The frontend properties object containing configuration like agentsNickname, etc.
  */
 export function useFrontendProperties(): Properties {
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useFrontendProperties.ts
```diff
@@ -12,31 +12,25 @@
...
-    refetchOnMountOrArgChange: 3600,
-    // Keep data fresh in cache even when component unmounts
-    refetchOnFocus: false,
-    refetchOnReconnect: false,
-  });
-
-  return useMemo(() => {
-    return frontendConfig?.frontend_settings?.properties || ({} as Properties);
-  }, [frontendConfig]);
+  return getConfig().properties;
```
florian-muller, line 35: We should deprecate `getConfig` and use `useFrontendProperties` with a `useGetFrontendConfigAgenticV1ConfigFrontendSettingsGetQuery` instead

florian-muller (changes requested):

florian-muller:
Closing, I created a simpler PR to fix the problem (not using deprecated `getConfig`) https://github.com/ThalesGroup/fred/pull/1611

PR #1614, opened 2026-05-09
Fix for the ToS accept page

marcfawaz:
## Summary

The user is not able to accept the ToS because the button is grayed. As a result, the user get stuck. This modificaiton enables the user to acceot the ToS.


dimitri-tombroff:
Hello Marc ! Sorry for this. I indeed saw all this and have fixed lots of problem i am also in a big (hopefully vertuous) phase where I improve a lote tha claude assistant to help us better generate code. This morning if you are there ok to have maybe a small sync up ?

PR #1626, opened 2026-05-18
feat: Add new pdf reader

Timothe-Le-Chatelier:
## Summary

Add a new processor to extract PDF to Markdown using the PyMuPDF4LLM library.

## Mandatory Checklist

- [ ] I followed [`docs/platform/DEVELOPER_CONTRACT.md`](../docs/platform/DEVELOPER_CONTRACT.md).

## Validation Notes

## Risk / Rollback


knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,273 @@
...
+            logger.info(
+                "[LITE_PDF][IMPLEM] Markitdown extraction | file=%s pages=%s chars=%s truncated=%s",
+                file_path.name,
+                result.page_count,
+                result.total_chars,
+                result.truncated,
+            )
+            return result
+        except Exception as e:
+            logger.warning(f"Markitdown PDF extraction failed, trying markitdown: {e}")
```
julien-buiss: wrong error message?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,273 @@
...
+        - If caller wants page control: use PyMuPDF (fast, deterministic).
+        - Else prefer markitdown (best quality/maintenance for whole-doc).
+        - On failures, log and fall back to the other path when possible.
+        """
+        opts = options or LiteMarkdownOptions()
+        result: LiteMarkdownResult = self._extract_pymupdf4llm(file_path, opts)
+
+        # Primary extraction: PyMuPDF4LLM
+        try:
+            result: LiteMarkdownResult = self._extract_pymupdf4llm(file_path, opts)
```
julien-buiss: why are we running the same function twice, once in the try/catch and once outside?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite_markdown_structures.py
```diff
@@ -88,3 +91,10 @@ def collapse_whitespace(text: str) -> str:
     while "\n\n\n" in t:
         t = t.replace("\n\n\n", "\n\n")
     return t
+
+
+def normalize_repeated_chars(text):
+    return re.sub(r'([.])\1{2,}', r'\1\1\1', text)
```
julien-buiss: only removes extra dots, is this intended?

julien-buiss (dismissed):

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
Timothe-Le-Chatelier: It's fixed!

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
Timothe-Le-Chatelier: It's fixed

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite_markdown_structures.py
Timothe-Le-Chatelier: To simplify the document summary.

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+
+    # ---- public API ----------------------------------------------------------
+
+    def extract(self, file_path: Path, options: LiteMarkdownOptions | None = None) -> LiteMarkdownResult:
+        """
+        Strategy:
+        - If caller wants page control: use PyMuPDF (fast, deterministic).
+        - Else prefer markitdown (best quality/maintenance for whole-doc).
+        - On failures, log and fall back to the other path when possible.
```
florian-muller, line 145: Not sure this docstring is still up to date ? It seems it always try PyMuPDF first and then fallback to Markitdown

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+                result.page_count,
+                result.total_chars,
+                result.truncated,
+            )
+            return result
+        except Exception as e:
+            logger.warning(f"Markitdown PDF extraction failed: {e}")
+
+        # Final fallback: return empty result (never return None)
+        return LiteMarkdownResult(document_name=file_path.name, page_count=0, total_chars=0, truncated=False, markdown="", pages=[], extras={})
```
florian-muller: I am not sure this is a good way of handling both extraction engine failing. It seems to me that `LitePdfMarkdownProcessor` is not checking for "empty" result and that the previous lite pdf extractor was just raising exception in case of failure. In addition, even if every caller of `extract` was checking for empty result (like `total_chars == 0`), I think we are loosing information (caller can't know how the extractor failed and can't catch it to handle it like it want).

If my diagnostic is not mistaken, I think extract here should raise an exception: raise again the exception of the last failed engine, or maybe better, raise a custom exception containing the 2 excetiion of from the 2 failed engine

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+
+            cleaned_pages.append(md_text)
+
+        md_text = "\n".join(cleaned_pages)
+
+        # Build final result in Markdown
+        return LiteMarkdownResult(
+            document_name=file_path.name,
+            page_count=len(pages),
+            total_chars=count_char - 1,
```
florian-muller, line 131: we could just do `len(md_text)` directly here. Is there an advantage to compute it ba summing multiple sub len result ?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+
+            # Apply your normalization pipeline
+            md_text = self._normalize_whitespace(text, opts)
+            md_text = self._normalize_repeated_chars(md_text, opts)
+            md_text = self._remove_additional_tags(md_text, opts)
+
+            count_char += len(md_text) + 1
+            if opts.max_chars and count_char > opts.max_chars:
+                truncated = True
+                continue
```
florian-muller: If `max_chars` is used (which I think its not ? Maybe we should remove it completly ?), the len of the last page will be added to `count_char` (so returned) but the content of the last page is not returned not appened. It result in a weird state (total_chars above max_char but actual page content lower than max_char).

I think it should also have a coherent behaviour with `enforce_max_chars`.

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+        md_text = "\n".join(cleaned_pages)
+
+        # Build final result in Markdown
+        return LiteMarkdownResult(
+            document_name=file_path.name,
+            page_count=len(pages),
+            total_chars=count_char - 1,
+            truncated=truncated,
+            markdown=md_text,
+            pages=[],
```
florian-muller: why not return the content by pages if we have them ? (if this is used, not sure about that. If that's not used, we should think about removing it)

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+        return remove_additional_tags(text) if opts.remove_additional_tags else text
+
+    def _extract_with_markitdown(self, file_path: Path, opts: LiteMarkdownOptions) -> LiteMarkdownResult:
+        if not self._md:
+            raise RuntimeError("markitdown not available for PDF conversion")
+
+        converted = self._md.convert(str(file_path))
+        # Pylance sometimes flags `.text`; guard with getattr for static peace of mind.
+        md = getattr(converted, "text", "") or ""
+        md = self._normalize_whitespace(md, opts)
```
florian-muller, line 72: why normalize only whitespace and not repeated chars and and tags ?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+    def _remove_additional_tags(self, text: str, opts: LiteMarkdownOptions) -> str:
+        return remove_additional_tags(text) if opts.remove_additional_tags else text
+
+    def _extract_with_markitdown(self, file_path: Path, opts: LiteMarkdownOptions) -> LiteMarkdownResult:
+        if not self._md:
+            raise RuntimeError("markitdown not available for PDF conversion")
+
+        converted = self._md.convert(str(file_path))
+        # Pylance sometimes flags `.text`; guard with getattr for static peace of mind.
+        md = getattr(converted, "text", "") or ""
```
florian-muller: That's weird, `MarkItDown.convert` return a type of `DocumentConverterResult` where I do not see a `text` property. It has a `markdown` property to get the text result and a deprecated `text_content` property. I don't see how this worked in the past (maybe it has nver worked) ?

(Also, as the result is typed, we can just use `converted.markdown` instead of convoluted `getattr`)

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite_markdown_structures.py
```diff
@@ -88,3 +91,12 @@ def collapse_whitespace(text: str) -> str:
...
     return t
+
+
+def normalize_repeated_chars(text):
+    # Replace any character repeated three or more times with exactly three occurrences of that character
+    return re.sub(r"(.)\1{2,}", r"\1\1\1", text)
+
+
+def remove_additional_tags(text):
+    return re.sub(r"\*\*==> picture .* intentionally omitted <==\*\*\n?", "", text)
```
florian-muller: Just by curiosity, where are these "==> picture .* intentionally omitted <==" coming from ? is it a pdf standard or is it some of our processor engine doing it ?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite_markdown_structures.py
```diff
@@ -88,3 +91,12 @@ def collapse_whitespace(text: str) -> str:
     while "\n\n\n" in t:
         t = t.replace("\n\n\n", "\n\n")
     return t
+
+
+def normalize_repeated_chars(text):
+    # Replace any character repeated three or more times with exactly three occurrences of that character
+    return re.sub(r"(.)\1{2,}", r"\1\1\1", text)
```
florian-muller: isn't it dangerous to apply this on other things that dot `.` ? Like if a doucment contain id (or zip code, version number...) with mutliple consecutive number like`id-00000123`, it could be important ?

I think we should only apply this to a subset of safe characters (dot, maybe dashes, underscore, equals...). We should be careful to not destroy markdown notation too (like title with equals `=====`)

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite_markdown_structures.py
```diff
@@ -39,8 +40,10 @@ class LiteMarkdownOptions:
     max_table_cols: int = 10
     include_images: bool = False
     page_range: Optional[Tuple[int, int]] = None  # inclusive (1-based for PDF)
-    max_chars: Optional[int] = 60_000
+    max_chars: Optional[int] = None
     normalize_whitespace: bool = True
+    normalize_repeated_chars: bool = True
+    remove_additional_tags: bool = True
```
florian-muller: I the 2 new tranformation added here but by adding this options in `LiteMarkdownOptions`, it give the impression that all processor who take a `LiteMarkdownOptions` can handle them (which is not true, only one handle them). 

If they are really specific to pdf or the engine used: we should maybe only refence them inside the processor

If they are generalist (which they seem to me), I think it would be coherent if they were applied to all the other processor (at least the ones taking a `LiteMarkdownOptions`). I see 2 way to do this:

- easy: we add a function that would take markdown text and `LiteMarkdownOptions` and apply all the needed transformation (ex: `def apply_normalization_pipeline(md: str, opts: LiteMarkdownOptions) -> str`). It would be more DRY, all processor use this function and we can change it in one place.

- or more refactoring (and less sure if this is a good idea): as this tranformation are made on markdown, we could run them outside of the `BaseLiteMdProcessor` abstraction. `BaseLiteMdProcessor` implementation are made to transform a specific format to markdown. Then we run markdown improvement pipeline on it, independently from the makrdown origin (which processor was used). But I don't know enough the processor code, I don't know if that's easy to do / a good idea.

What do you think about it ?

knowledge-flow-backend/knowledge_flow_backend/core/processors/input/lightweight_markdown_processor/lite2_pdf_to_md_processor.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+            return result
+        except Exception as e:
+            logger.warning(f"Markitdown PDF extraction failed: {e}")
+
+        # Final fallback: return empty result (never return None)
+        return LiteMarkdownResult(document_name=file_path.name, page_count=0, total_chars=0, truncated=False, markdown="", pages=[], extras={})
+
+
+class LitePdfMarkdownProcessor(BaseMarkdownProcessor):
```
florian-muller, line 179: This class looks identical to the old `LitePdfMarkdownProcessor` (except for some added `doc.close()`), this create lots of duplicated code. 
 
Just to be sure, the idea is to remove the old `lite_pdf_to_md_processor.py` soon ? If yes, we could maybe add a comment in `lite_pdf_to_md_processor.py` about this, saying this code is deprecated and should be removed soon if the new version works ?

knowledge-flow-backend/config/configuration_worker.yaml
```diff
@@ -52,7 +52,7 @@ processing:
         do_ocr: false
       input_processors:
         - suffix: ".pdf"
-          class_path: knowledge_flow_backend.core.processors.input.lightweight_markdown_processor.lite_pdf_to_md_processor.LitePdfMarkdownProcessor
+          class_path: knowledge_flow_backend.core.processors.input.lightweight_markdown_processor.lite2_pdf_to_md_processor.LitePdfMarkdownProcessor
```
florian-muller, line 55: If you want `lite2_pdf_to_md_processor` to be used by default in integration/production... you should also change the values in the chart (line 37 of `deploy/charts/fred/values.yaml`) !

florian-muller (commented):
Greate first contribution, gg ! The new `LitePdfToMdExtractor` is really clean, looking forward to see this in production :)

simcariou (approved):

PR #1656, opened 2026-05-27
1654 mermaid rendering hardening and markdown first UI contract

dimitri-tombroff:
This PR make the mermaid and code extracts well supported in the chat UI. The specifications have been updated. Fred design is aligned with openai and claude, not a perfect protocol because mermaid or code sample is still embedded ito text, requiring UI functions toi parse and render correctly. 

This compromise and design is defined in specifications and explained.


simcariou:
@dimitri-tombroff , don't merge yet. I need to give you a demo, or for you to test it on your end. I've started some work that I think is quite nice with CodeBlock streaming. We can take a look on Monday when you get back, or you can test it in the meantime and let me know :)

PR #1676, opened 2026-06-03
fix: resolve dual Alembic head and extend migration CI to swift branch

florian-muller:

**ID:** no ID — mechanical fix because the migration head divergence was an accidental branching error, not a planned feature.

**Problem in one sentence:** `655ea83030fa_add_default_prompt_usage` was created branching off `d6e7f8a9b0c1`, skipping the `f0f1e2d3c4b5` → `ee5a5b163f34` → `d4e5f6a7b8c9` chain, leaving two Alembic heads and breaking `make db-check-heads`.

**Backlog ref:** n/a — untracked incident fix.

---

**RFC written?** not required — mechanical fix: the correct linear ordering of migrations is unambiguous from the commit dates and chain structure.

**Plan presented to the team before implementation?** Diagnosed and confirmed fix inline with the developer.

**Confirmation received?** Developer confirmed the fix approach before implementation.

---

## 3. What you built

- Updated `down_revision` in `655ea83030fa_add_default_prompt_usage.py` from `d6e7f8a9b0c1` to `f0f1e2d3c4b5`, restoring a single linear head.
- Added `swift` to `push.branches` and `pull_request.branches` in `Check-migrations.yml` so the CI gate catches this class of divergence before it reaches `swift` or `main`.

---

## 4. Proof of quality

**Tests added or updated:** none needed — `make db-check-heads` is the verification and it now passes.

| n/a  | n/a  | checked via `make db-check-heads` which outputs "Single head confirmed." |

**`make code-quality` output:** n/a — no Python logic changed, only migration metadata.

**`make test` output:** n/a — migration head check is the relevant test, passes locally.

---

## 5. Docs updated

| --- | --- |
| Backlog `[ ]` item is now done | n/a — not tracked |
| New behaviour, API field, or contract change | n/a |
| Frozen contract touched | n/a |
| UX component implemented or visual status changed | n/a |
| WORKPLAN sprint item finished | n/a |

---

## 6. Close-out statement

```
## Task close-out
- Code: updated down_revision in 655ea83030fa to point to f0f1e2d3c4b5; added swift to CI trigger
- Tests: make db-check-heads passes — "Single head confirmed."
- Docs updated: none — mechanical fix
- Backlog: none — not tracked
- Skipped steps: RFC (mechanical fix, no design decision); GitHub issue (developer confirmed inline)
```

---

## 7. Risk and rollback

**What breaks if this is wrong?** Migrations would run in the wrong order; the `default_prompt_usage` table would be created before the GCU nullable columns fix, which is safe — no dependency between them.

**How do we roll back?** Revert `down_revision` back to `d6e7f8a9b0c1` in the migration file.


florian-muller:
Closing in favour of #1677

PR #1725, opened 2026-06-11
feat: JSON schema generation and CI validation for backend configs

florian-muller:
## Summary
- Add JSON schema generation and CI validation for backend configs
- Fix type errors: `secret_key` in `MinioStorageConfig`, `MinioPullSource`, and `MinioFilesystemConfig` was typed as `Optional[str]` but the `model_validator` always injects `MINIO_SECRET_KEY` and raises if absent — so `None` is never a valid post-construction value; corrected to `str`

## Test plan
- [ ] CI passes (`make type-check` in `knowledge-flow-backend`)
- [ ] JSON schema generation workflow passes
- [ ] MinIO-based storage, pull source, and filesystem config still load correctly from env

### DIFF
--- a/apps/fred-agents/config/configuration_prod.yaml
+++ b/apps/fred-agents/config/configuration_prod.yaml
@@ -28,7 +28,7 @@ security:
     api_url: "http://localhost:9080"
 
 ai:
-  knowledge_flow_url: "http://localhost:9528/knowledge-flow/v1"
+  knowledge_flow_url: "http://localhost:8111/knowledge-flow/v1"
   timeout:
     connect: 20
     read: 60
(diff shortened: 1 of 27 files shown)


apps/fred-agents/config/configuration_prod.yaml
```diff
@@ -27,7 +28,7 @@ security:
     api_url: "http://localhost:9080"
 
 ai:
-  knowledge_flow_url: "http://localhost:8111/knowledge-flow/v1"
+  knowledge_flow_url: "http://localhost:9528/knowledge-flow/v1"
```
simcariou: Oupsie

simcariou (changes requested):

apps/fred-agents/config/configuration_prod.yaml
florian-muller: ho, good catch ! Thanks

Issue #1753, opened 2026-06-16
kea: robust durable file-ingestion UX in library folders


dimitri-tombroff:
## Problem

Ingesting files into a library folder is fragile. The UI uploads files **one request
per file, sequentially**, each holding a streaming HTTP connection open for up to
**30 minutes** inside a modal bound to the upload drawer. All progress lives in React
state that is **wiped on navigation/close**. If the user leaves or reloads, progress is
lost; with hundreds of files, file N+1 doesn't even start until file N finishes.

## What this changes

Submit fast, then let the **library rows themselves** reflect durable backend status.
The metadata store (already exposing per-stage `processing.stages`) is the source of
truth — no held connection, no remembered `workflow_id` in the client.

### Backend (`knowledge-flow-backend`)
- **New `POST /knowledge-flow/v1/schedule-documents`** — persists each document (so it
  appears in the library immediately) and submits the Temporal workflow, then returns
  right away. The existing `/upload-process-documents` (streaming) endpoint is unchanged.
- **New `POST /knowledge-flow/v1/documents/processing/reconcile`** — lists a folder's
  documents and reconciles still-pending ones against Temporal (see Robustness below).
- `Processing.workflow_id` added (optional, stored in the existing JSON metadata blob →
  **no DB migration**), written by the API **before** submit and threaded through the
  schedulers so it can't race the worker.

### Frontend
- "Process" uploads now go through `/schedule-documents` in **bounded batches of 20**;
  the drawer closes immediately and the library refreshes.
- The library **polls** while the open folder has non-terminal documents (3s, auto-stops),
  so status updates live and survives navigation/reload.
- Per-row status indicator: **Pending / Processing / Failed** as a chip; **Ready** shows
  the familiar **R/P/V/S/M** stage pills inline (completed stages use the theme `--success`
  green, matching the row's search icon).

## Robustness (corner cases)

A document must never stay "pending in fred" while no live workflow will advance it. No
fred-side timeout/retry logic was added — we only reflect Temporal's own verdict.

- **Temporal unreachable at schedule time** (`submit` throws) → documents are durably
  marked **Failed** (not left pending) and reported per-file.
- **Worker flooded** → eventually consistent (calmly shows Processing).
- **Worker down past the workflow timeout** → reconciliation reads Temporal's status
  (a client `describe`, works with the worker down) and marks **Failed** durably.
- **Temporal server unreachable on read** → left as-is (never false-fail on a transient
  outage).
[...]

PR #1754, opened 2026-06-16
#1753 fix the weak file ingestion UI

dimitri-tombroff:
ONLY temporary not to be merged in swift already much better. This is an improvement for kea only.


simcariou:
Je teste sur le k3d là!

simcariou (approved):

PR #1790, opened 2026-06-19
new version of pdf extractor

Timothe-Le-Chatelier:
# Pull Request

> **Read this before you type anything.**
>
> Opening a PR is the _last_ step, not the first. If you designed this alone,
> implemented it without sharing the plan, and are now hoping the review will
> validate the approach — stop. Close this tab. Go read
> [`docs/swift/platform/DEVELOPER_CONTRACT.md`](../docs/swift/platform/DEVELOPER_CONTRACT.md)
> and come back when the steps below are already done.
>
> This is not process theater. Every question below exists because something
> went wrong when it was skipped. Fill them in honestly. A short honest answer
> beats a long evasive one every time.

---

**ID:** 

**Problem in one sentence:**

**Backlog ref:** 

---

This section cannot be skipped. If the answer to any question is "no" or
"I didn't", explain why — do not leave it blank.

**RFC written?**

**Plan presented to the team before implementation?**

**Confirmation received?**

---

## 3. What you built

---

## 4. Proof of quality

**Tests added or updated:**

|      |      |                |

**`make code-quality` output (paste the last line or "all checks passed"):**

```

```

**Raw `basedpyright` output (required if a touched package keeps a non-empty baseline file):**

```

```

**`make test` output (paste the summary line — pass count, 0 failures):**

```

```

If you touched multiple packages, paste one block per package.

---

## 5. Docs updated

Work through this line by line. If an item does not apply, write "n/a" and say why.

| What changed                                                      | File updated |
| ----------------------------------------------------------------- | ------------ |
| Backlog `[ ]` item is now done                                    |              |
| New behaviour, API field, or contract change                      |              |
| Frozen contract touched (`execution.py`, `agent_app.py`, OpenAPI) |              |
| UX component implemented or visual status changed                 |              |
| Phase progress row exists for this area                           |              |
| WORKPLAN sprint item finished                                     |              |
| Code and a design doc now diverge                                 |              |

---

## 6. Close-out statement

```
## Task close-out
- Code:
- Tests:
- Docs updated:
- Backlog:
- Skipped steps:
```

---

## 7. Risk and rollback

**How do we roll back?**


dimitri-tombroff (approved):

dimitri-tombroff:
Nice cleanup, @Timothe-Le-Chatelier — the pluggable base_pdf_extractor + docling/pymupdf split reads really well, and the config churn is consistent end-to-end (YAML configs, JSON schema, PdfPipelineConfig, Helm values, Dockerfile all agree — extra="forbid" makes that easy to trust). Net −1138 lines is a great direction. 👍

Approving / merging — a few small things to pick up, ideally a quick look at #1 before or right after merge since it's the only possible regression:

Dockerfile leftover. This PR drops ENV DOCLING_ARTIFACTS_PATH=/app/docling-artifacts, but it looks like it's still referenced lower down in Dockerfile-prod (mkdir -p … ${DOCLING_ARTIFACTS_PATH} / chown -R … ${DOCLING_ARTIFACTS_PATH}), where it now expands to empty. Worth removing those leftover usages.

One bad image shouldn't sink the whole doc. _use_image_describer re-raises as RuntimeError, which unwinds up to convert_file_to_markdown and silently downgrades the entire document to the markitdown fallback. A single unreadable image then loses all the good structured output — a per-image try/except that logs and continues would be safer.

Logging nit (pre-existing, easy while you're here). logger.warning("… Failed to use ocr model : %s | Error : ", ocr_model.backend_name, e) passes two args but has only one %s, so the exception never makes it into the message.

Declare the transitive deps you import directly. cv2/numpy/shapely (in images_feature_extraction.py) and paddlex (in download_models.py) are only present via paddleocr today — worth adding them to pyproject.toml explicitly so a future paddleocr bump can't quietly drop them.

Optional / pre-existing: processors are cached as singletons but self.folder is created once in __init__ and rmtree'd every call, so a second document on the same instance writes into a deleted dir. Not introduced here, but the rewrite was a natural moment to switch to a fresh tempdir per call.

Tests look solid for the dispatch + OCR/VLM substitution paths; if you want to tighten later, the pymupdf extractor and the unknown-extractor fallback aren't directly covered yet. Thanks for this — great refactor. 🚀

Issue #1645, opened 2026-05-22
Maintain Table Formatting in Markdown Files


marcfawaz:
As a technical agent auther,
I want the ingestion process (light, medium and rich) to keep the tables in the markdown files as a whole, 
so that the formatting and data integrity are preserved.

Context / Business Value:
Markdown files often contain tables that are crucial for presenting data in a structured manner. Ensuring that these tables are ingested as a whole maintains the integrity and readability of the documentation. This is particularly important for technical writers who rely on accurate and well-formatted tables to convey complex information effectively.

Acceptance Criteria:
Given a markdown file with a table,
When the ingestion process runs,
Then the table is preserved as a whole in the output.

dimitri-tombroff:
Good catch marc.

Issue #1774, opened 2026-06-18
CHAT-12: human-friendly tool call labels in the chat trace


simcariou:
## Task

**ID:** CHAT-12
**Track:** chat UI rendering (`ThoughtTrace`) — extends RUNTIME-05 tool-call thought synthesis
**RFC:** [`docs/swift/rfc/TOOL-DISPLAY-NAME-RFC.md`](docs/swift/rfc/TOOL-DISPLAY-NAME-RFC.md)
**Backlog:** [`docs/swift/backlog/CHAT-UI-BACKLOG.md` §12](docs/swift/backlog/CHAT-UI-BACKLOG.md)

## Problem

In the thought/tool trace, tool calls still surface raw technical identifiers. RUNTIME-05 now wraps each tool call in a `tool_use` thought titled `"Calling <name>"`, but:

- the standalone `tool_call` row still shows the raw `name` + `name(args)` (`traceUtils.ts` `entryLabel` / `primaryTextForEntry`);
- the friendly thought and the raw row are redundant for the same call;
- the humanizer `_tool_thought_title` (`react_runtime.py`) is crude — `mcp__tavily__web_search` becomes `"Calling mcp  tavily  web search"`.

## Scope — Phase 1 (contract-free)

Decisions (2026-06-18): **merge** the raw tool_call row into its `tool_use` thought; **defer** the authored `display_name` contract field to Phase 2.

**Backend (`libs/fred-runtime`)**
- [ ] Improve `_tool_thought_title()` in `react_runtime.py`: strip `mcp__<provider>__` prefix, collapse whitespace, sentence-case, optional common-tool dictionary.
- [ ] Update `tests/test_react_thinking.py` for the new strings (incl. MCP-prefixed case).

**Frontend (`apps/frontend`)**
- [ ] Extend `groupTraceEntries()` (`traceUtils.ts`) to pair a `tool_use` thought with its adjacent `tool_call` (+ result by `call_id`) and suppress the standalone combo row.
- [ ] Attach `args` + result to the merged thought entry so `TraceDetailDrawer` (`traceDrawerContext`) shows the raw technical payload.
- [ ] Humanize the orphan-combo fallback (tool_call with no preceding `tool_use` thought) with a helper mirroring `_tool_thought_title`.

**Verification**
- [ ] Backend + frontend unit tests; `make code-quality` + `make test` in `libs/fred-runtime` and `apps/frontend`.
[...]

PR #1815, opened 2026-06-24
Fix for Maintain Table Formatting in Markdown Files #1645

pdubey28-sketch:
# Pull Request — Fix for Maintain Table Formatting in Markdown Files (ThalesGroup#1645)

**Branch:** `pdubey28-sketch:Issue#1645` → `pdubey28-sketch:swift`
**Commit:** `c8be3eb6066118b76bc2c2ae3d1d0f0308658e8d` (single commit, 4 files, +256 / −26)

---

**ID:** _no ID — mechanical fix._ GitHub issue [`ThalesGroup/fred#1645`](https://github.com/ThalesGroup/fred/issues/1645) is the source of truth; no entry exists in `docs/swift/data/id-legend.yaml`. If tracking is wanted, propose adding a `RUNTIME-xx` row (adapter dedup/sort) and a knowledge-flow ingestion row before merge.

**Problem in one sentence:** Markdown tables ingested into the RAG store were returned to the LLM out of document order and with broken header/row structure, so row-level questions over tables (e.g. "what is the default of `property25`?") returned wrong or empty answers.

**Backlog ref:** none — no `[ ]` item existed for this bug. Recommend adding one before merge.

---

**RFC written?** Not required — mechanical fix. The change restores the documented "preserve tables" behaviour of `SemanticSplitter` for `.md`/`.txt` inputs and repairs a half-applied chunk-ordering fix in the v2 runtime adapter. No new contract or endpoint; one optional internal DTO field (`VectorSearchHit.chunk_index`) that the adapter was already trying to read.

**Plan presented to the team before implementation?** No — the bug was diagnosed and patched in a single debugging session from a reproducible failure (`large-table-test_pd1.md` returning tables in similarity-score order with repeated headers). The plan is captured retroactively in `docs/MARKDOWN_TABLE_INGESTION_FIX.md`.

**Confirmation received?** No prior confirmation — this PR is the request for review.

---

## 3. What you built

One commit (`c8be3eb`) with two coordinated fixes for the same regression:

1. **Ingestion — `SemanticSplitter`.** Added `_auto_annotate_unmarked_tables` so plain-markdown and text inputs get the same `<!-- TABLE_START -->` / `<!-- TABLE_END -->` wrapping that `PdfMarkdownProcessor` and `DocxMarkdownProcessor` already emit (detection is conservative: a pipe row must be immediately followed by a markdown separator row; fenced code blocks and quoted blocks are skipped). Tables that fit in `chunk_size` are emitted as a single atomic chunk. Oversized tables fall back to `_split_large_table`, which splits only at row boundaries and repeats the header + separator in every part, stamping `block_type="markdown_table"`, `table_id`, `table_part`, `row_start`, `row_end`.
2. **Retrieval — `VectorSearchHit` + `_to_hit` + `_invoke_knowledge_search`.** Added the missing `chunk_index: Optional[int] = None` field to `VectorSearchHit`, populated it from doc metadata in `VectorSearchService._to_hit()`, and added a `prev_content_is_table` guard to the dedup loop in the v2-runtime adapter so the first table chunk keeps its header even when an intro/text chunk from the same document precedes it. The sort by `chunk_index` in `_invoke_knowledge_search` was already in place but silently raising `AttributeError`; it now works.

### Files changed

| File                                                                                                                     | Lines       |
| ------------------------------------------------------------------------------------------------------------------------ | ----------- |
| `apps/knowledge-flow-backend/knowledge_flow_backend/core/processors/output/vectorization_processor/semantic_splitter.py` | +204 / −26 |
| `apps/knowledge-flow-backend/knowledge_flow_backend/features/vector_search/vector_search_service.py`                     | +4 / −0    |
| `libs/fred-core/fred_core/store/vector_search.py`                                                                        | +1 / −0    |
| `libs/fred-runtime/fred_runtime/integrations/v2_runtime/adapters.py`                                                     | +47 / −0   |

---

## 4. Proof of quality

**Tests added or updated:**

| Test                                                                | File                                                                                                           | What it covers                                                                            |
| ------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
| `test_small_table_remains_in_one_chunk`                             | `apps/knowledge-flow-backend/tests/processors/output/vectorization_processor/test_semantic_splitter_tables.py` | Small table stays whole                                                                   |
| `test_text_before_and_after_table_chunked_separately`               | same file                                                                                                      | Prose chunks normally; table whole; no marker leakage                                     |
| `test_large_table_splits_only_on_row_boundaries_and_repeats_header` | same file                                                                                                      | Row-boundary splits only; header+separator repeated in every part; rows in original order |
| `test_split_preserves_row_order_in_metadata`                        | same file                                                                                                      | `row_start`/`row_end` contiguous and monotonic                                            |
| `test_multiple_tables_are_each_preserved_independently`             | same file                                                                                                      | Distinct ids per table, both preserved                                                    |
| `test_pipe_text_without_separator_row_is_not_treated_as_table`      | same file                                                                                                      | Pipe prose without separator is left as prose                                             |
| `test_fenced_code_block_with_pipes_is_not_treated_as_table`         | same file                                                                                                      | Code-fence pipes not misdetected                                                          |
| `test_auto_annotation_does_not_misdetect_separator_inside_text`     | same file                                                                                                      | Separator-like prose without pipes not misdetected                                        |
| `test_pre_annotated_tables_are_idempotent`                          | same file                                                                                                      | Existing PDF/DOCX annotations keep their ids and content                                  |

Retrieval-side change is covered by 5 manual assertions in `C:\tmp\probe_fix_verify.py` (all pass). No formal pytest exists yet for `FredKnowledgeSearchToolInvoker._invoke_knowledge_search` (would require mocking the vector-search HTTP client) — recommended as a follow-up.

```
TODO — paste output from:
  (cd apps/knowledge-flow-backend && make code-quality)
  (cd libs/fred-core && make code-quality)
  (cd libs/fred-runtime && make code-quality)
```

**Raw `basedpyright` output (required if a touched package keeps a non-empty baseline file):**

```
TODO — run basedpyright for each touched package and paste output
if a baseline file is present.
```

**`make test` output:**

```
test_semantic_splitter_tables.py — 9/9 PASS

TODO — paste `make test` summary line for:
  apps/knowledge-flow-backend
  libs/fred-core
  libs/fred-runtime
```

---

## 5. Docs updated

| What changed                                                      | File updated                                                                               |
| ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------ |
| Backlog`[ ]` item is now done                                     | n/a — no backlog item existed for issue #1645                                             |
| New behaviour, API field, or contract change                      | `docs/MARKDOWN_TABLE_INGESTION_FIX.md` (new) — splitter behaviour + auto-annotation rules |
| Frozen contract touched (`execution.py`, `agent_app.py`, OpenAPI) | n/a —`VectorSearchHit.chunk_index` is an optional internal DTO field                      |
| UX component implemented or visual status changed                 | n/a                                                                                        |
| Phase progress row exists for this area                           | n/a                                                                                        |
| WORKPLAN sprint item finished                                     | n/a                                                                                        |
| Code and a design doc now diverge                                 | n/a                                                                                        |

---

## 6. Close-out statement

```
## Task close-out
- Code:
    - SemanticSplitter: auto-annotate unmarked markdown tables; atomic
      table chunk when it fits; row-boundary split with repeated
      header+separator when oversized
    - VectorSearchHit: added optional `chunk_index` field
    - VectorSearchService._to_hit: pass `chunk_index` from metadata
    - v2_runtime adapter `_invoke_knowledge_search`: dedup guard so the
      first table chunk after an intro chunk keeps its header
- Tests: 9 new ingestion tests in
    apps/knowledge-flow-backend/tests/processors/output/vectorization_processor/test_semantic_splitter_tables.py
    (all pass); 5 probe assertions for the retrieval-side change (pass);
    no formal adapter pytest added (follow-up).
- Docs updated: docs/MARKDOWN_TABLE_INGESTION_FIX.md (new)
- Backlog: none — no [ ] item existed; recommend adding one before merge
- Skipped steps: Step 1 (RFC) — mechanical fix; Step 2 (backlog entry)
    — no pre-existing item; Step 3 (developer confirmation) — diagnosed
    and patched from a reproducible failure in one session, this PR is
    the request for confirmation; Step 3.5 (GitHub issue) — #1645 exists.
```

---

## 7. Risk and rollback

- **Splitter false positives.** If `_auto_annotate_unmarked_tables` misclassifies pipe-using prose as a table, that block becomes a single atomic chunk instead of being prose-split. Detection is conservative (pipe row must be immediately followed by a markdown separator row of only `|`, `-`, `:`, whitespace) and covered by `test_pipe_text_without_separator_row_is_not_treated_as_table` and `test_fenced_code_block_with_pipes_is_not_treated_as_table`.
- **Old documents without `chunk_index`.** The adapter sort falls back to `999_999` for hits missing `chunk_index`, so legacy-indexed documents sort to the end of the result list. Acceptable degradation; re-ingestion restores correct ordering.
- **Dedup over-stripping.** The `prev_content_is_table` guard strips a header only when the immediately preceding same-document chunk was also a table chunk. Worst case: a continuation chunk keeps a duplicate header (cosmetic), not a wrong answer.

**How do we roll back?**

Revert commit `c8be3eb` (the four touched files + the new test file). The splitter change is process-local — restart `knowledge-flow-backend` after revert. Already-ingested documents do not need re-ingestion to roll back (the new metadata fields are additive), but tables ingested under the new splitter will revert to the previous broken behaviour on next retrieval.

### DIFF
--- a/apps/knowledge-flow-backend/knowledge_flow_backend/core/processors/output/vectorization_processor/semantic_splitter.py
+++ b/apps/knowledge-flow-backend/knowledge_flow_backend/core/processors/output/vectorization_processor/semantic_splitter.py
@@ -25,6 +25,27 @@ logger = logging.getLogger(__name__)
 
 _WS_CLASS = r"[ \t\r\n\u00A0]+"  # space, tabs, CR/LF, NBSP
 
+# Marker shape used to protect tables across the pipeline. The PDF/DOCX
+# processors emit numeric ids; here we use an "auto_" prefix so we never
+# collide with annotations produced upstream.
+_TABLE_START_RE = re.compile(r"^\s*<!--\s*TABLE_START:id=([^\s>]+?)\s*-->\s*$")
+_TABLE_END_RE = re.compile(r"^\s*<!--\s*TABLE_END\s*-->\s*$")
+_FENCE_RE = re.compile(r"^\s*(```|~~~)")
+
+
+def _is_pipe_row(line: str) -> bool:
+    s = line.strip()
+    return "|" in s and not s.startswith("```") and not s.startswith("~~~")
+
+
+def _is_pipe_separator(line: str) -> bool:
+    """Markdown table separator: only `|`, `-`, `:`, whitespace; at least one of each
+    `|` and `-`."""
+    s = line.strip()
+    if "|" not in s or "-" not in s:
+        return False
+    return all(ch in "|-: \t" for ch in s)
+
 
 def _short(s: str, n: int = 160) -> str:
     """Short, visible snippet for logs (make whitespace visible)."""
@@ -62,6 +83,91 @@ class SemanticSplitter(BaseTextSplitter):
         self.chunk_overlap = chunk_overlap
         self.preserve_tables = preserve_tables
 
+    def _auto_annotate_unmarked_tables(self, text: str) -> str:
+        """
+        Detect plain (unannotated) Markdown pipe tables and wrap them with
+        ``<!-- TABLE_START:id=auto_N --> ... <!-- TABLE_END -->`` markers so the
+        rest of the splitter pipeline treats them as atomic blocks.
+
+        Why:
+            Issue #1645 — table rows must not be split across chunks. Upstream
+            processors (DOCX, PDF) already annotate; ``.md`` / ``.txt`` and the
+            lightweight processors do not. Doing the detection here means light,
+            medium and rich ingestion modes all benefit without per-processor
+            duplication.
+
+        Detection rules (intentionally conservative):
+            - skip pre-existing TABLE_START/TABLE_END blocks untouched
+            - skip lines inside fenced code blocks (``` or ~~~)
+            - a table starts on a pipe-row whose next non-empty line is a
+              Markdown separator row (only ``|``, ``-``, ``:``, whitespace and
+              at least one of ``|`` and ``-``)
+            - the table extends through subsequent contiguous pipe-rows; a
+              blank line ends it
+        """
+        lines = text.splitlines()
+        out: list[str] = []
+        i = 0
+        auto_idx = 0
+        in_fence = False
+        fence_token: str | None = None
+
+        while i < len(lines):
+            line = lines[i]
+            stripped = line.lstrip()
+
+            # Track fenced code blocks; pipes inside them are never tables.
+            if _FENCE_RE.match(line):
+                token = stripped[:3]
+                if not in_fence:
+                    in_fence = True
+                    fence_token = token
+                elif token == fence_token:
+                    in_fence = False
+                    fence_token = None
+                out.append(line)
+                i += 1
+                continue
+            if in_fence:
+                out.append(line)
+                i += 1
+                continue
+
+            # Preserve existing annotations untouched.
+            if _TABLE_START_RE.match(line):
+                out.append(line)
+                i += 1
+                while i < len(lines):
+                    out.append(lines[i])
+                    if _TABLE_END_RE.match(lines[i]):
+                        i += 1
+                        break
+                    i += 1
+                continue
+
+            # Pipe table: a pipe-row immediately followed by a separator row.
+            if i + 1 < len(lines) and _is_pipe_row(line) and _is_pipe_separator(lines[i + 1]):
+                start = i
+                i += 2  # consume header + separator
+                while i < len(lines) and lines[i].strip() and _is_pipe_row(lines[i]):
+                    i += 1
+                table_md = "\n".join(lines[start:i]).rstrip()
+                if table_md:
+                    out.append(f"<!-- TABLE_START:id=auto_{auto_idx} -->")
+                    out.append(table_md)
+                    out.append("<!-- TABLE_END -->")
+                    auto_idx += 1
+                continue
+
+            out.append(line)
+            i += 1
+
+        # Preserve a single trailing newline if the original text had one.
+        result = "\n".join(out)
+        if text.endswith("\n") and not result.endswith("\n"):
+            result += "\n"
+        return result
+
     def _extract_and_replace_tables(self, text: str) -> Tuple[str, dict]:
         """
         Extracts annotated Markdown tables from the text and replaces them with unique placeholders.
@@ -93,7 +199,9 @@ class SemanticSplitter(BaseTextSplitter):
         """
         Splits a large Markdown table into smaller chunks based on the configured chunk size.
 
-        Preserves the table header in each chunk and adds metadata including table ID and chunk index.
+        Preserves the table header + separator row in **every** chunk and stamps
+        metadata so consumers can re-stitch the table or filter on
+        ``block_type=='markdown_table'``. Row order is preserved.
 
         Args:
             table_md (str): The Markdown string of the full table.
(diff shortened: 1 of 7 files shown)


dimitri-tombroff:
Hi @pdubey28-sketch 👋 — thanks a lot for this, and especially for the really thorough write-up. The document-order fix and the row-boundary table splitting are exactly the right approach, and the conservative detection rules are nicely done. 🙌

Just two things I'd love your help understanding before we merge — both questions, not blockers:

1. The preserve_tables flag. In the previous reinsert logic, preserve_tables=True kept a large table as a single atomic chunk, and only False allowed it to be split. The new code always splits when the table is larger than chunk_size, so the flag isn't consulted in that path anymore. Was that an intentional simplification (i.e. we always want row-boundary splitting now)? If so, all good — could you just confirm, so we can note it? If not, we may want to keep honoring the flag.

2. Table detection vs. header dedup. Your _is_pipe_row detection nicely handles tables without a leading pipe, but the dedup in the runtime adapter decides "is this a table chunk?" with hit.content.startswith("|"). For a leading-pipe-less table, that check would miss it and the repeated header wouldn't get stripped. Is that case in scope, or are we assuming all ingested tables are normalized to leading-pipe form? Related: could you confirm chunk_index is actually populated in the chunk metadata for these table sub-chunks? The splitter stamps table_part / table_chunk_id, and I want to make sure the new chunk_index sort in the adapter isn't quietly falling back to the 999_999 default for them.

No rush on any of this — happy to pair on it if that's easier. Thanks again for the great contribution! 🚀

pdubey28-sketch:
@dimitri-tombroff Please find my answers inline

1. preserve_tables Flag (Dead Code)
preserve_tables is no longer used by the semantic chunking logic, making it dead code after the #1645 refactor.
Regardless of the flag value:
Small tables (≤ chunk_size) are always kept as a single chunk.
Large tables are always split at row boundaries using _split_large_table().
Since all production and test callers already use the default (True), there is no functional regression.

2. 
Q1 – Leading Pipe (|) Table Detection Mismatch
A pre-existing inconsistency exists for raw Markdown tables without a leading |.
Table detection accepts any row containing |, but deduplication only recognizes chunks starting with |, so repeated headers may not be removed.
This is not related to Issue #1774 and does not affect current tests (large-table-test_pd1.md) or normal ingestion, as DOCX/PDF/CSV processors already normalize tables to the standard leading-pipe format.
Recommendation: Track this as a separate issue and normalize leading pipes during _auto_annotate_unmarked_tables instead of changing the runtime dedup logic.

Q2 – chunk_index Population
chunk_index is correctly populated during vectorization, not during table splitting.
It survives metadata sanitization and is stored in OpenSearch, making it available for retrieval sorting.
Temporary metadata (table_part, table_chunk_id) is intentionally discarded before storage.
The 999999 fallback is only used for:
Old indexes created before chunk_index was stored (requires re-indexing), or
Unit tests that bypass the vectorization pipeline.
For all freshly indexed documents, including table sub-chunks, chunk_index is present and the fallback is never used.

simcariou:
merged on another updated branch, Thanks!

PR #1819, opened 2026-06-24
feat(CHAT-13): GeoJSON map rendering in chat (Leaflet)

### LINKED ISSUE #1818: CHAT-13: Render GeoJSON fences as interactive Leaflet maps in chat
**Task ID:** CHAT-13              
                                                                                                                                                               
  **RFC:** docs/swift/rfc/CHAT-GEOJSON-MAP-RENDERING-RFC.md 
                                                                                                                                       
  **Backlog:** docs/swift/backlog/CHAT-UI-BACKLOG.md §14                                                                                                                                           
                                                                                                                                                                                                   
  
## Problem

Agent replies frequently contain geographic data as a GeoJSON `FeatureCollection`.
Today that data renders as a raw ` ```json ` code block: a wall of coordinates the
user has to read by hand. The chat renderer already turns two other fenced payloads
into rich visualizations — ` ```mermaid ` → flowchart (`MermaidBlock`) and
` ```mindmap-json ` → interactive mindmap (`MindMapBlock`) — but there is no
equivalent for maps.

We want GeoJSON to render as an interactive map (Leaflet), consistent with the
existing mermaid/mindmap treatment, and we want this to be a **renderer capability
available to every agent**, not a one-off for the test assistant's hardcoded
`markdown` scenario.

### 1.1 Why not the `ui_parts` / `GeoPart` path

The SDK already defines a `GeoPart` (`libs/fred-sdk/fred_sdk/contracts/context.py`)
that an agent can emit in `ui_parts` on the `final` event. An earlier attempt at
this feature rendered maps from that structured part. It was rejected for this use
case because:

- It only fires when an agent deliberately emits a `GeoPart`. The actual content
  users see today — including the test assistant's `markdown` scenario — embeds
  GeoJSON **inside the markdown text** as a ` ```json ` fence, which the `ui_parts`
  path never touches.
- It requires every agent to adopt a new emission pattern to get maps, whereas
  mermaid/mindmap "just work" from markdown content.

`STREAMING-RENDER-GUARD-RFC.md §1.3` explicitly scoped GeoJSON out, on the
assumption it would always arrive as a `GeoPart`. This RFC revisits that: GeoJSON
in a markdown fence is the common case and should render as a map.

The `GeoPart`/`ui_parts` path remains valid and is **not removed** — it is simply
not the mechanism used here.

---

## 2. Proposed solution

### 2.1 Principle

Add one branch to the `MarkdownRenderer` fenced-code dispatch, alongside the
existing `mindmap`/`mermaid` branches, that routes GeoJSON to a new `GeoMapBlock`
molecule rendering an interactive Leaflet map.

### 2.2 Detection rule

| Fence | Rendered as | Rule |
|---|---|---|
| ` ```geojson ` | `GeoMapBlock` | Always — explicit opt-in |
| ` ```json ` | `GeoMapBlock` | Only when the body parses as a GeoJSON `FeatureCollection` |
| ` ```json ` (non-geo) | `CodeBlock` | Falls through unchanged |

The `json` case is intentionally conservative: it parses the fence and checks
`type === "FeatureCollection"`. Anything else (plain JSON, a bare `Feature`,
invalid JSON) falls through to the normal `CodeBlock`, so no existing behavior
regresses. The detector (`isGeoJsonFeatureCollection`) is a pure function exported
from `GeoMapBlock` and is unit-testable without a DOM.

### 2.3 `GeoMapBlock` molecule

Mirrors `MindMapBlock`'s shape — a self-contained molecule taking
`{ code, language }`, parsing the fence, and rendering. Stack:
`react-leaflet` (`MapContainer` + OpenStreetMap `TileLayer` + `GeoJSON`).
Dependencies (`leaflet`, `react-leaflet`, `@types/leaflet`) are already in
`package.json`; this is their first consumer.

Behavioral decisions:

- **Points → SVG pin via `L.divIcon`.** Leaflet's default PNG marker icons
  (`marker-icon.png` etc.) do not survive Vite bundling and render as broken
  images. An inline-SVG teardrop pin has no external asset and always renders.
  (A `circleMarker` variant was also tried; the SVG pin matches the familiar
  Leaflet look from leafletjs.com and is unambiguous to users.)
- **Polygons / lines** are styled from the feature's `color` / `fillOpacity`
  properties, falling back to the theme `--primary` token.
- **Popups** bind the feature's `name` / `title` property.
- **Auto-fit** to the data's bounds via a small `FitBounds` child (`useMap` +
  `L.geoJSON(data).getBounds()`), falling back to a central-Europe view when the
  bounds are not resolvable.

### 2.4 Layout / stacking

Two CSS rules in `GeoMapBlock.module.css`:

- **Landscape aspect** (`height: 20rem`, `width: 100%`) so the map is a wide
  rectangle like leafletjs.com, not a tall square.
- **Own stacking context** (`position: relative; z-index: 0`) on the Leaflet
  container. Leaflet assigns internal pane/control z-indices up to 1000; without
  a containing stacking context these escape to page level and overlap the
  floating chat input bar (`.inputOverlay`, `z-index: 1`) on scroll. Confining
  them to a `z-index: 0` context keeps the whole map below the input bar.

---

## 3. Alternatives considered

### 3.1 Structured `GeoPart` in `ui_parts` (rejected for this use case)

Render maps from an agent-emitted `GeoPart` on the `final` event, plumbed through
`ThreadMessage` → `AssistantTurn`. **Rejected:** does not render GeoJSON that lives
in markdown text (the common case), and forces every agent to adopt a new emission
pattern. Not removed from the platform — just not the mechanism here. See §1.1.

### 3.2 Leaflet default PNG markers (rejected)

Use `L.Icon.Default` with the bundled `leaflet/dist/images/*.png` URLs.
**Rejected:** still rendered as broken images in the Vite build (a well-known
Leaflet+bundler asset-path problem). The inline-SVG `divIcon` removes the external
asset dependency entirely.

### 3.3 `circleMarker` for points (rejected)

Draw points as SVG circles. **Rejected:** works and needs no assets, but reads as
a faint dot rather than a recognizable location marker; the SVG pin is clearer and
matches user expectation from the official Leaflet page.

### 3.4 Render all ` ```json ` fences as maps (rejected)

**Rejected:** would break every non-geographic JSON snippet. The
`FeatureCollection` discriminator keeps detection safe.
                                                                                                                                                                                                 
  ## Proposal                                                                                                                                                                                      
  Add a MarkdownRenderer branch routing ```geojson (and ```json that parses as a                                                                                                                   
  FeatureCollection) to a new GeoMapBlock (react-leaflet). Renderer-level, so every                                                                                                                
  agent benefits. Frontend-only, no contract change.

Vanshajbehl:
Closes #1818. Implements RFC docs/swift/rfc/CHAT-GEOJSON-MAP-RENDERING-RFC.md (CHAT-13). Frontend-only; no contract change. Validated with make -C apps/frontend code-quality."

### DIFF
--- a/apps/frontend/src/rework/components/shared/molecules/GeoMapBlock/GeoMapBlock.module.css
+++ b/apps/frontend/src/rework/components/shared/molecules/GeoMapBlock/GeoMapBlock.module.css
@@ -1,0 +1,77 @@
+.block {
+  margin: var(--spacing-s) 0;
+  border: calc(var(--spacing-3xs) / 4) solid var(--outline-muted);
+  border-radius: var(--radius-s);
+  background: linear-gradient(180deg, var(--surface-container-low), var(--surface-container-lowest));
+  overflow: hidden;
+  color: var(--on-surface);
+}
+
+.header {
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-s);
+  border-bottom: calc(var(--spacing-3xs) / 4) solid var(--outline-muted);
+  background: color-mix(in srgb, var(--surface-container) 88%, transparent);
+  padding: var(--spacing-2xs) var(--spacing-s);
+  min-height: calc(var(--spacing-l) + var(--spacing-m));
+  color: var(--on-surface);
+}
+
+.lang {
+  color: var(--on-surface-muted);
+  font: var(--font-label-small);
+  font-family: monospace;
+  letter-spacing: 0.05em;
+  text-transform: uppercase;
+}
+
+.title {
+  color: var(--on-surface);
+  font: var(--font-label-large);
+}
+
+.content {
+  display: block;
+  padding: var(--spacing-s);
+}
+
+.map {
+  /* position + integer z-index establishes a stacking context so Leaflet's
+     internal pane/control z-indices (up to 1000) stay confined below the
+     floating chat input bar instead of overlapping it on scroll. */
+  position: relative;
+  z-index: 0;
+  border: calc(var(--spacing-3xs) / 4) solid var(--outline-muted);
+  border-radius: var(--radius-s);
+  background: var(--surface-container-lowest);
+  width: 100%;
+  /* Landscape rectangle (like leafletjs.com) rather than a tall square box. */
+  height: 20rem;
+}
+
+/* divIcon wrapper: strip Leaflet's default white box so only the SVG pin shows. */
+.pin {
+  border: none;
+  background: transparent;
+}
+
+.error {
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-s);
+  background: var(--error-container);
+  padding: var(--spacing-m);
+}
+
+.errorLabel {
+  margin: 0;
+  color: var(--on-error-container);
+  font: var(--font-label-large);
+}
+
+.errorText {
+  margin: 0;
+  color: var(--on-error-container);
+  font: var(--font-body-medium);
+}
--- a/apps/frontend/src/rework/components/shared/molecules/GeoMapBlock/GeoMapBlock.tsx
+++ b/apps/frontend/src/rework/components/shared/molecules/GeoMapBlock/GeoMapBlock.tsx
@@ -35,6 +35,11 @@ interface GeoMapBlockProps {
 const FALLBACK_CENTER: [number, number] = [50, 8];
 const FALLBACK_ZOOM = 4;
 
+// GeoJSON properties are agent-supplied and may be untrusted (e.g. RAG content).
+// The `color` property flows into SVG markup and Leaflet path styles, so we only
+// accept hex / rgb(a) / hsl(a) values and fall back to the theme accent otherwise.
+const SAFE_COLOR = /^#[0-9a-f]{3,8}$|^(rgb|hsl)a?\([\d.,%\s/]+\)$/i;
+
 // Self-contained SVG pin rendered via L.divIcon. We deliberately avoid Leaflet's
 // default PNG marker icons: their image paths don't survive Vite bundling and show
 // as broken images. An inline SVG has no external asset, so it always renders.
@@ -125,9 +130,12 @@ export function GeoMapBlock({ code, language = "geojson" }: GeoMapBlockProps) {
     );
   }
 
+  // Validate an agent-supplied color against the allowlist; fall back to accent.
+  const safeColor = (c: unknown): string => (typeof c === "string" && SAFE_COLOR.test(c.trim()) ? c.trim() : accent);
+
   const styleFn = (feature?: Feature<Geometry>): PathOptions => {
     const props = (feature?.properties ?? {}) as Record<string, unknown>;
-    const color = (props.color as string) || accent;
+    const color = safeColor(props.color);
     return {
       color,
       weight: 2,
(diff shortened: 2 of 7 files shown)


dimitri-tombroff:
Hi @Vanshajbehl — Thank you! The feature is well thought out, and I want to call out that you followed the full project workflow — RFC, backlog entry, id-legend row, and the PMO-board mirror all included. That's genuinely rare for a first contribution and very much appreciated. 🙌

The map rendering itself is clean (nice touch using an inline-SVG divIcon to dodge the Leaflet+Vite marker-asset problem, and confirming the Leaflet deps were already in package.json). I just have two small security fixes I'd like before we merge — both quick, both in GeoMapBlock.tsx.

The context: GeoJSON properties come from the agent's reply, which can carry untrusted content (e.g. RAG documents), and two spots currently render those values as HTML, which opens an XSS hole.

1. Popup label → rendered as HTML. layer.bindPopup(String(label)) treats the string as HTML, so a feature like {"name": ""} would execute. Using a text node escapes it safely:


const onEachFeature = (feature: Feature<Geometry>, layer: Layer) => {
  const props = (feature?.properties ?? {}) as Record<string, unknown>;
  const label = props.name ?? props.title;
  if (label != null && label !== "") {
    const el = document.createElement("span");
    el.textContent = String(label);   // textContent escapes everything
    layer.bindPopup(el);
  }
};
2. color property → interpolated into the SVG pin markup. In buildPinIcon, an attacker-controlled props.color flows straight into the html template (e.g. "#000\"/></svg>), so it's worth validating against a simple CSS-color allowlist and falling back to the theme accent otherwise:


const SAFE_COLOR = /^#[0-9a-f]{3,8}$|^(rgb|hsl)a?\([\d.,%\s/]+\)$/i;
const safeColor = (c: unknown) =>
  typeof c === "string" && SAFE_COLOR.test(c.trim()) ? c.trim() : accent;
Then use const color = safeColor(props.color); in both pointToLayer and styleFn instead of (props.color as string) || accent.

Both are tiny (~15 lines total), and a quick manual check with a malicious name/color in a test FeatureCollection should confirm they're closed.

One optional, non-blocking thought for later: the map pulls tiles from the public OpenStreetMap server, which sends viewport + client IP to a third party — for enterprise/offline deployments it'd be nice to make the tile URL configurable down the line. Totally fine as a follow-up, not something for this PR.

Thanks again — this is a great contribution and very close. 🚀

dimitri-tombroff (changes requested):
Marking as request-changes — see my earlier comment above for the two small sanitization fixes (popup label + pin color). Both are quick and the PR is very close. Thanks again! 🙏

simcariou:
Also you can test it using the test agent (no LLM) in the fred-agents backend

You can send "markdown" in the chat to see a geomap type json be rendered in the chat. Or if it does not work (not properly parsed), maybe consider updating the bug markdown string in the test agent's code  in `graph_steps.py` to display a proper json that works with your code

dimitri-tombroff:
Hi @Vanshajbehl — closing the loop here, sorry for the delay.

After sitting with it a bit longer, I don't think we should merge this one as-is. The SDK already has a typed path for exactly this (`GeoPart` on `ui_parts`), already wired end-to-end via the `geo.render_points` tool, and an earlier RFC (`STREAMING-RENDER-GUARD-RFC.md`) explicitly scoped GeoJSON out of the markdown pipeline on the assumption it would always arrive that way. Rendering from a ```json fence works, but it re-opens that assumption and duplicates a path that already exists.

That said, the map itself is genuinely good work and I didn't want it to go to waste — I've ported the Leaflet setup (the divIcon pins, and both sanitization fixes you added — popup as a text node, the color allowlist) into `GeoPartRenderer` so it renders straight from the typed `GeoPart` instead. It's landing internally under CHAT-13, credited to you in the commit message.

Thanks again for the thorough job on this, RFC and all — that made it easy to build on. Sorry it didn't land the way you originally shipped it, and feel free to close this PR whenever suits you.

PR #1821, opened 2026-06-24
feat(RUNTIME): add deep agent dispatch + Report Generator with web search

abhishek04071:
## Summary
- **Fix dispatch bug**: `DeepAgentDefinition` was silently running as `ReActRuntime` because it inherits `ReActAgentDefinition`. Added `elif isinstance(definition, DeepAgentDefinition)` branch before the ReAct fallthrough so `DeepAgentRuntime` is correctly selected.
- **Add `ReportGeneratorDefinition`**: a `DeepAgentDefinition` that researches a topic and writes a structured cited report. Appears automatically in the Create Agent menu under the `deep` category pill — no frontend change needed.
- **DuckDuckGo web search tool**: free, no API key required. Registered as an inprocess MCP provider (`mcp-web-search-google`) so the agent can search the web before writing reports.
- **Add `inprocess_toolkit_factory` parameter to `create_agent_app`**: allows pods to register custom in-process MCP tool providers.
- **Fix `ExecutionCategory` enum in `agenticOpenApi.ts`**: `"deep"` was missing, causing type errors when the control-plane returned a deep agent category.

## Test plan
- [ ] Go to Create Agent → see **Report Generator** card with `deep` category pill
- [ ] Create an instance, enable **Google Web Search** tool in the Tools tab
- [ ] Ask: `generate a report on cybersecurity threats` — agent searches web then writes structured report

### DIFF
--- a/apps/fred-agents/config/mcp_catalog.yaml
+++ b/apps/fred-agents/config/mcp_catalog.yaml
@@ -242,6 +242,14 @@ servers:
     provider: "web_github_readonly"
     auth_mode: "no_token"
 
+  - id: "mcp-web-search-google"
+    name: "Google Web Search"
+    description: "Search the web using Google Custom Search API."
+    enabled: true
+    transport: "inprocess"
+    provider: "web_search_google"
+    auth_mode: "no_token"
+
   - id: "mcp-knowledge-flow-statistics"
     name: "mcp-knowledge-flow-statistics"
     description: "mcp.servers.statistics.description"
--- a/apps/fred-agents/config/mcp_catalog.yaml
+++ b/apps/fred-agents/config/mcp_catalog.yaml
@@ -242,12 +242,12 @@ servers:
     provider: "web_github_readonly"
     auth_mode: "no_token"
 
-  - id: "mcp-web-search-google"
-    name: "Google Web Search"
-    description: "Search the web using Google Custom Search API."
+  - id: "mcp-web-search"
+    name: "Web Search"
+    description: "Search the web using DuckDuckGo."
     enabled: true
     transport: "inprocess"
-    provider: "web_search_google"
+    provider: "web_search"
     auth_mode: "no_token"
 
   - id: "mcp-knowledge-flow-statistics"
--- a/apps/fred-agents/fred_agents/main.py
+++ b/apps/fred-agents/fred_agents/main.py
@@ -30,6 +30,7 @@ from fastapi import FastAPI
 from fred_runtime.app import AgentPodConfig, create_agent_app, load_agent_pod_config
 
 from fred_agents.registry import REGISTRY
+from fred_agents.web_search import inprocess_toolkit_factory
 
 
 def create_app(config: AgentPodConfig | None = None) -> FastAPI:
@@ -52,6 +53,7 @@ def create_app(config: AgentPodConfig | None = None) -> FastAPI:
     return create_agent_app(
         registry=REGISTRY,
         config=resolved_config,
+        inprocess_toolkit_factory=inprocess_toolkit_factory,
     )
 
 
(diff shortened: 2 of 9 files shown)


simcariou (changes requested):
If I may, since the web_search tool is independent of the actual backend so maybe do not write Google everywhere ?

FYI if you omit the backend parameter in ddgs.text(...) it defaults to "auto" so maybe google, maybe not google.
> https://pypi.org/project/ddgs/#user-content-1-text

abhishek04071:
@simcariou : Thanks for the feedback! Fixed — renamed mcp-web-search-google → mcp-web-search, provider key web_search_google → web_search, and GoogleSearchToolkit → WebSearchToolkit. No more Google references in the naming. The backend stays ddgs with its default auto behavior.

apps/fred-agents/config/mcp_catalog.yaml
```diff
@@ -242,6 +242,14 @@ servers:
     provider: "web_github_readonly"
     auth_mode: "no_token"
 
+  - id: "mcp-web-search"
+    name: "Web Search"
+    description: "Search the web using DuckDuckGo."
```
simcariou, line 247:
```suggestion
    description: "Search some information on the internet."
```
 Since ddgs is not really DuckDuckGo

simcariou (changes requested):
Also after having a second look I have some design suggestions:
- I find the location of web_search.py a bit weird (straight in fred_agents) I'd imagine a more suitable location for example in one of the libs/ folders where you would define the tools
- I would have put inprocess_toolkit_factory in a separate file since it is not about web_search. web_search is part of the toolkit but when we have more (inevitably in the near future) it makes no sense to have it in here. Maybe create another file specifically for the toolkit where we add more in process tools below web_search in another "if" condition.

I'm saying to put this in one of the libs folder so other agent pods (not the prepackaged agents from fred-agents) can benefit from this tool as well

I'm in my phone so that's not really practical but I'll have a look tomorrow morning at work to give you a better review. In the meantime if that makes sense, you can try to take into consideration the couple of things I said 

Cheers,

simcariou:
@abhishek04071 
look for:

```
from fred_sdk.support.builtins import (
    TOOL_REF_ARTIFACTS_PUBLISH_TEXT,
    TOOL_REF_ATTACHMENTS_READ_IMAGE,
    TOOL_REF_GEO_RENDER_POINTS,
    TOOL_REF_KNOWLEDGE_SEARCH,
    TOOL_REF_LOGS_QUERY,
    TOOL_REF_RESOURCES_FETCH_TEXT,
    TOOL_REF_TRACES_SUMMARIZE_CONVERSATION,
)
```

to see the pattern we have for adhoc tools (adhoc = non mcp) :)

dimitri-tombroff:
Thank you for this contribution — the DeepAgent dispatch fix and the `ReportGeneratorDefinition` example are genuinely useful, and `web_search` is a good idea for a first Deep agent.

We're closing this without merging as-is, for two reasons:

1. `inprocess_toolkit_factory` (the mechanism `web_search.py` adds) duplicates work that landed on `swift` in the meantime — `build_inprocess_toolkit` (MIGR-03.03, `fred_runtime/integrations/inprocess_toolkit_registry.py`), with a different signature. That's the source of the current merge conflict, and matches what @simcariou already flagged: file location and design should follow the existing adhoc-tool pattern rather than adding a second toolkit-factory mechanism.
2. Since this PR was opened, `swift` landed its capability framework (manifest + middleware) as the standard way to add new tool integrations. A new tool like `web_search` is a better fit there than as a raw MCP-catalog/inprocess-toolkit entry — and the runtime dispatch code this PR touches has moved since, so a rebase would be needed regardless.

We're picking up the DeepAgentRuntime observability gap this PR surfaced (Deep silently bypassing the platform's audit/KPI middleware) separately. Thanks again for surfacing this — happy to point you at the capability authoring guide if you'd like to take another pass at `web_search` on top of it.

Issue #1975, opened 2026-07-10
Agent suspension lifecycle for broken or revoked capabilities


florian-muller:
## Parent

#1961 — Agent Capability RFC (`CAPAB-01`, RFC §3.9)

## What to build

A broken capability **suspends the agent — it never silently degrades** (an agent missing its tools would confidently answer from priors; that trust failure is worse than unavailability).

`suspended` is a platform-forced state distinct from the editor's `disabled` toggle, carrying a typed reason: `capability_unavailable`, `capability_access_revoked`, or `capability_config_invalid`.

- **Lazy detection:** the pod hitting a mismatch at agent assembly returns a typed error naming the capability; control-plane flips the instance to suspended.
- **Proactive detection:** a low-priority reconciliation sweep on the control-plane lifecycle Temporal queue runs whenever aggregated manifests change (pod deploy/registration) or an enablement tuple is deleted — agents leave the catalog before anyone hits an error.
- **Product behavior:** suspended agents hidden from chat-only team members; editors/owners see the agent with a warning and a locked enable toggle; the edit form renders the broken capability in an error state with plain-language text and the two fix paths (untick the capability and re-save, or contact a platform admin). A successful save clears the suspension — no second mechanism.
- **Observability:** structured log + metric per suspension (the per-capability health column joins the Tier 3 admin dashboard later).

## Acceptance criteria

- [ ] Assembly-time mismatch → typed error → instance suspended with the correct reason
- [ ] Reconciliation sweep suspends affected instances on manifest change / enablement revocation, before a user hits an error
- [ ] Suspended agents hidden from chat-only members; editors see warning + locked enable toggle
- [ ] Edit-form error state with both fix paths; unticking the capability and saving clears the suspension
- [ ] `upgrade_config` failure surfaces as `capability_config_invalid` with the reset-and-save message
- [ ] Structured log + metric emitted per suspension

## Blocked by

- #1974 (informed by the mismatch findings of the spike #1971)

florian-muller:
[...]

PR #2005, opened 2026-07-17
feat(CAPAB-01): resting capability health + revoke-impact preview, fix grant-side revive

florian-muller:
## What & why

Two linked gaps in the capability suspension lifecycle (#1975), sharing one derived predicate.

### 🐛 Fix — grants never un-suspended their agents

`disable_capability_for_team` suspends dependents inline, but the three **grant** paths — enable-for-team, reset-to-default-on, and the default-on toggle — bare-returned `0`. The only path that clears an availability suspension was the reconciliation sweep, which has **no scheduled host yet** (the Temporal lifecycle queue is named as the intended one but is unwired).

**Consequence:** an admin who disabled a capability, watched N agents suspend, then re-enabled it, left those N agents suspended **indefinitely**. The undo didn't undo.

`revive_dependent_instances` now runs after each grant. It passes the **real** availability facts (`can_use` ∩ pod manifests), not the revoke path's synthetic `selected - {id}` — a grant *can't* fake that set, because a second selected capability may still be broken. It clears only availability reasons; a `capability_config_invalid` suspension is left untouched (RFC §3.9, only a successful save clears it).

### ✨ Feature — resting health column + impact preview

- The dashboard's **health column** read session-only React state that reset on refresh. It now shows the real **at-rest** count from a derived predicate (`selected ∩ not-usable`), computed once per list request (one ReBAC `ListObjects` per team, **not** per agent).
- **Attribution is derived, never read from `suspension_reason`** (which records *why*, not *which*). An agent broken by `capa1` while also selecting `capa2` is **not** miscounted against `capa2` — the exact miscount a naive `suspension_reason IS NOT NULL AND capa2 IN selected` query produces.
- **Unreachable pods report as UNKNOWN, never broken**, matching the sweep's `skipped_unreachable` rule — a pod restart is not an outage.
- A new `GET /admin/capabilities/{id}/revoke-impact` endpoint powers a live **"this will suspend N agents"** preview in the pre-disable confirmation dialog, with an affected-agent drill-down (name + team).

## Scope note

Personal-space capability scope (from `feat/agent-capability-1961`) is **intentionally not included**. Those two commits predate AUTHZ-05's authorization-model rewrite (`admin`→`platform_admin`, `owner/manager/member`→`team_*`) that landed on `swift-golive`, so rebasing them is a **port**, not a merge — it deserves its own PR and review. This work is team-scoped only.

## Testing

- **Backend:** 428 tests pass. New `test_capability_impact.py` covers the multi-capability attribution correctness, the revive fix, the config-invalid safety guard (caught a real counter bug during development), and unreachable-pod handling.
- **Frontend:** `make code-quality` clean (tsc + prettier); 45 CapabilitiesPage tests pass (1 updated for the new health states, 2 added).
- Generated OpenAPI client regenerated via `make update-control-plane-api`; new route added to `authz-endpoint-matrix.yaml`.

### DIFF
--- a/apps/control-plane-backend/control_plane_backend/capabilities/api.py
+++ b/apps/control-plane-backend/control_plane_backend/capabilities/api.py
@@ -38,6 +38,7 @@ from control_plane_backend.capabilities.enablement import (
 from control_plane_backend.capabilities.schemas import (
     CapabilityDefaultOnResult,
     CapabilityEnablementList,
+    CapabilityImpactPreview,
     EnableTeamCapabilityRequest,
     SetCapabilityDefaultOnRequest,
     TeamCapabilityEnablementResult,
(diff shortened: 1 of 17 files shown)


dimitri-tombroff:
Florian same info here this one is already incorporated into the consolidation branch i am working on. Keep you informed here.

PR #2007, opened 2026-07-17
feat(CAPAB-01): resting capability health + revoke-impact preview, fix grant-side revive

florian-muller:
## What & why

Two linked gaps in the capability suspension lifecycle (#1975), sharing one derived predicate.

### 🐛 Fix — grants never un-suspended their agents

`disable_capability_for_team` suspends dependents inline, but the three **grant** paths — enable-for-team, reset-to-default-on, and the default-on toggle — bare-returned `0`. The only path that clears an availability suspension was the reconciliation sweep, which has **no scheduled host yet** (the Temporal lifecycle queue is named as the intended one but is unwired).

**Consequence:** an admin who disabled a capability, watched N agents suspend, then re-enabled it, left those N agents suspended **indefinitely**. The undo didn't undo.

`revive_dependent_instances` now runs after each grant. It passes the **real** availability facts (`can_use` ∩ pod manifests), not the revoke path's synthetic `selected - {id}` — a grant *can't* fake that set, because a second selected capability may still be broken. It clears only availability reasons; a `capability_config_invalid` suspension is left untouched (RFC §3.9, only a successful save clears it).

### ✨ Feature — resting health column + impact preview

- The dashboard's **health column** read session-only React state that reset on refresh. It now shows the real **at-rest** count from a derived predicate (`selected ∩ not-usable`), computed once per list request (one ReBAC `ListObjects` per team, **not** per agent).
- **Attribution is derived, never read from `suspension_reason`** (which records *why*, not *which*). An agent broken by `capa1` while also selecting `capa2` is **not** miscounted against `capa2`.
- **Unreachable pods report as UNKNOWN, never broken**, matching the sweep's `skipped_unreachable` rule — a pod restart is not an outage.
- A new `GET /admin/capabilities/{id}/revoke-impact` endpoint powers a live **"this will suspend N agents"** preview in the pre-disable confirmation dialog, with an affected-agent drill-down (name + team).

### ✨ Feature — health column drill-down by team

The health column's **"N suspended"** count named nothing — an admin saw *that* agents were broken but not *which*, or in *which team* to go fix access. It's now a button that opens a drawer listing the suspended agents **grouped by team**.

- The named instances ride inline on the aggregate list item (`suspended_instance_details`) instead of behind a new endpoint: the resting-impact walk already visits every instance, so naming them is a list append and the drawer opens with **no extra request**.
- Reusing `revoke-impact` for this was rejected — it *excludes already-broken agents* by design, so its list would be empty for exactly the capabilities showing suspensions. The drill-down and the count come from the same derivation, so they always agree.
- The count is now a keyboard-reachable button (focus ring + `aria-label`), not a static span; the drawer resolves opaque team ids to names and sorts teams and agents stably.

## Scope note

Personal-space capability scope (from `feat/agent-capability-1961`) is **intentionally not included** — it predates AUTHZ-05's authorization-model rewrite and needs a separate port. This work is team-scoped only.

## Testing

- **Backend:** capability + authz-matrix suites pass on the `swift` base. `test_capability_impact.py` covers multi-capability attribution correctness, the revive fix, the config-invalid safety guard, unreachable-pod handling, and the new `collect_instances` drill-down naming (id + team).
- **Frontend:** `make code-quality` clean (tsc + prettier).
- Generated OpenAPI client regenerated via `make update-control-plane-api`; new route added to `authz-endpoint-matrix.yaml`.

### DIFF
--- a/apps/control-plane-backend/control_plane_backend/capabilities/api.py
+++ b/apps/control-plane-backend/control_plane_backend/capabilities/api.py
@@ -38,6 +38,7 @@ from control_plane_backend.capabilities.enablement import (
 from control_plane_backend.capabilities.schemas import (
     CapabilityDefaultOnResult,
     CapabilityEnablementList,
+    CapabilityImpactPreview,
     EnableTeamCapabilityRequest,
     SetCapabilityDefaultOnRequest,
     TeamCapabilityEnablementResult,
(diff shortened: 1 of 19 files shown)


dimitri-tombroff:
Florian for info, i took this and already incorporated it into a consolidation branch i am deploying for monday morning so hopefully monday morning it is included already. I will tell you here what branch that is (still woring on it)

Issue #2023, opened 2026-07-20
Swift UX bug pass: personal prompts in team space, library tree picker in agent form, capability toggle layout shift, empty-folder dropdowns


simcariou:
Grouped UX bug pass from live testing (2026-07-20). Four related frontend/UX bugs, fixed together in one branch.

## Bugs

1. **Personal prompts visible in team space (hide them).** `GET /teams/{team_id}/prompts/context` intentionally merges personal prompts into team context (PROMPTS.md design); product decision reversed — personal prompts must only appear in the personal space. Fix in `list_context_prompts` (single point covering the chat composer picker and the agent-form prompt picker) + PROMPTS.md update.

2. **Agent form: library/tag selection is a raw text TagInput — restore the tree picker.** `TuningFieldRenderer` renders `document_access`'s `library_tag_ids`/`document_uids` config fields as a generic `TagInput` (manual tag typing). The tree picker (`DocumentLibraryScopePicker`) already exists but is only used in the chat composer. Fix: additive `ui.widget` hint on `UIHints` (fred-sdk), set on those fields in the `document_access` manifest, dispatched in `TuningFieldRenderer`. Also add the legacy tool's missing "RAG scope" config field to `document_access` for config parity with "Document search (legacy)". (Backend tools port itself stays tracked by #1906.)

3. **Capability activation layout shift.** Toggling a capability switches the name from `--font-label-medium` to `--font-title-small` (`CapabilityCard.module.css`), changing header height and jumping the whole list. Fix: keep font metrics identical on activation (weight/color change only).

4. **Resources/Agents tabs: expanding an empty folder shows an empty dropdown.** The corpus workspace already has the reference empty-state pattern (`rework.resources.empty.folder` hint in `DocumentWorkspace.tsx`). Apply the same explanatory empty message to `AgentFilesystemBrowser` and the My space/Team space filesystem browser.

## Related
- "Created by <uid>" name resolution is NOT here — tracked by #1952 (implemented in the same branch).
- Prompt category reduction (7 categories, drop monitoring/migration defaults) is a #2010 checklist item, also implemented in the same branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

simcariou:
Implemented on branch `swift` (working tree, pending commit) — all four bugs plus the two tracked siblings (#1952 full scope, prompt-category reduction from #2010's checklist), in one pass.

**What landed**
[...]

PR #2033, opened 2026-07-20
feat: document_access config/chat parity with the legacy search tool

simcariou:
## Summary

Follow-up to the #2023 UX pass: the Document access capability now offers the same agent-configuration surface and chat composer controls as "Document search (legacy)", so agents can move off the legacy MCP path without losing any UI affordance.

- **New config toggles** (all default on): File attachments, Search policy picker in chat, RAG scope picker in chat + Default RAG scope select — alongside the existing scope-picker toggle. Library binding was already covered by `library_tag_ids` (now a tree picker), which also pins the chat scope picker read-only.
- **`chat_controls()` emits the same stock widgets** as the MCP capability (`attach_files`, `document_scope`, `search_policy`, `rag_scope`). Picked values travel on `RuntimeContext`, which the v2 document-search adapter already honors (per-turn `search_policy` fallback, `general_only` corpus skip) — the pickers are functional, not cosmetic. Zero frontend change needed.
- **Precedence**: with the search-policy picker shown, the configured policy is only the picker's *default* (the tool passes `None` so the per-turn value wins); with the picker hidden it is enforced as-is.
- **Manifest bumped to 0.2.0** — additive with defaults: stored 0.1.0 slices revalidate unchanged, and the control-plane chat-controls cache invalidates on the version key.
- **Legacy tool**: the "Bound document libraries" raw tag-id input now renders as the library tree via `ui.widget: document_libraries` in the pod's `mcp_catalog.yaml`.

Requires a fred-agents pod restart to advertise the 0.2.0 manifest and the updated catalog.

## Test plan

- New/updated fred-runtime tests: default parity set + ordering, per-toggle hiding, config values becoming picker defaults, search-policy precedence (shown → port gets `None`; hidden → enforced).
- fred-runtime suite: 557 passed; `make code-quality` clean.


simcariou:
Added `acd13bd5`: new additive `UIHints.visible_when` hint — a field renders only while the referenced sibling's effective value is truthy (display-only; stored value kept; backends keep their own gating). First consumer: the legacy tool's "Bound document libraries" tree now only shows when "Bind to specific libraries" is checked (`mcp_catalog.yaml`). Both OpenAPI clients regenerated, contract §8.18 amended. Suites re-run: fred-sdk 233, control-plane 476, frontend tsc/prettier clean.

simcariou:
Added `1d1a1732`: document_access now mirrors the legacy tool's **exact** configuration surface — split "Document library picker" / "Document picker" toggles, "Bind to specific libraries" gating the bound-libraries tree (`ui.visible_when`), bound ids inert while unbound (both at chat_controls and search time). Manifest 0.3.0 with a before-validator so stored older slices revalidate unchanged (single scope toggle maps onto the split ones; a pre-existing library scope stays binding). fred-runtime 559 passed, quality clean.

Issue #46, opened 2025-06-23
Add mechanisms to allow LLM to rephrase what the user want when searching for documents


florian-muller:
Allow the agent making the document research to better phrase what he need to search, maybe in multiples times ?

Example:
- User ask a question badly, using wrong term.
- Agent search bad term given by user: it gets document of mid quality but better understand what the user is talking about
- Agent think a bit and found a better search
- Agent use the search tool a second time and retrieve good document
- Give a good response

PR #2162, opened 2026-07-29
eval ui polish 6 fixes

### LINKED ISSUE #2141: Evaluation UI/backend polish: 6 fixes from a walkthrough (rerun button, JSON export, list scale, analysis service, run metadata, settings back nav)
## Context

Walkthrough of the evaluation UI (Fred settings — `/settings/evaluations`) done together with Dimitri. Six issues found across the frontend (fred, `apps/frontend/src/rework/.../TeamSettingsEvaluations/`) and the backend (`fred-agent-evaluator`, `apps/fred-evaluation-backend/fred_evaluation_backend/`). Grouped here as one tracking issue; can be split into sub-issues if the fixes don't land together.

---

### 1. Redundant header "Rerun" button on the runs list

The `Evaluations > <evaluation>` runs page has a "↺ Rerun · Instance · …" button in the header that reruns the most recent run (`EvaluationRuns.tsx:263-290`, `handleRerunRun(mostRecentRun)`). Every row in the runs table now has its own per-row "Rerun" button (`EvaluationRuns.tsx:387-392`) that does the exact same thing for any run. The header button is redundant duplication.

**Fix**: remove the header rerun button (keep "+ New run"), or repurpose it into something that isn't already covered by the per-row action.

### 2. No typed/versioned JSON export, only clipboard copy

The case detail drawer (`EvaluationRunDetail.tsx`, `CaseDetail` component) has a copy-to-clipboard button (`handleCopy`, line 533-546) that does `navigator.clipboard.writeText(JSON.stringify(caseData, null, 2))` — the raw generated `EvaluationCaseResponse` payload, no schema/version wrapper, no file download.

**Fix**: add a download action that produces a clear, typed, versioned JSON file (e.g. `{ schema_version, ... }`) suitable for keeping/sharing with an AI agent for discussion. Open question to resolve during implementation: case-level only, or also a run-level export (all cases + aggregated scores)? Should the analysis result (see #4) be included when available?

### 3. No scale beyond a handful of items (evaluations list + runs list)

- `Evaluations.tsx:35` — the evaluations list (`useListEvaluationsQuery({ teamId })`) loads everything, no pagination/sort/search.
- `EvaluationRuns.tsx:76` — the runs-of-one-evaluation list has a hardcoded `limit: 200`, no server-side pagination beyond that, no search, and the table isn't explicitly sorted (only `mostRecentRun` is recomputed separately for the rerun button, see #1).

At 250 evaluations or 250 runs this UI breaks down or becomes unusable.

**Fix**: add search-by-name, explicit sort (most recent / oldest), and real pagination on both the evaluations list and the runs list. Check whether `fred-agent-evaluator`'s list endpoints already support offset/limit/sort/query params, or whether that needs adding backend-side first.

### 4. "Analysis service not configured" is undocumented, and backlog doc is stale

- `runs/api.py:359-363` (fred-agent-evaluator) — `analyze_run` returns 503 "Analysis service not configured" when `AnalysisClient(configuration.analysis)` failed to build at startup (`main.py:90-98`), which happens silently (just a `logger.warning`) whenever the analysis provider/model isn't configured for the deployment.
- `docs/backlog/EVAL-LLM-ANALYSIS-BACKLOG.md` (fred-agent-evaluator) still says `Status: [ ] not started`, even though the `/runs/{id}/analyze` endpoint and the frontend "Analyze" button are both implemented.

**Fix**:
- Document how to configure `configuration.analysis` (provider/model) for a deployment that wants the Analyze feature.
- Update `EVAL-LLM-ANALYSIS-BACKLOG.md` to reflect the real implementation state.
- Once available, the analysis result should be part of the exported JSON dump (see #2).

### 5. Run information panel: missing author, misleading "Scoring profile" field

- **Missing author**: `EvaluationRunRow.created_by` is stored in the DB (`runs/models.py:62`) and passed into `start_run()`, but the `EvaluationRun` response schema (`runs/schemas.py:84-108`) never exposes it, so the frontend "Run information" panel (`EvaluationRunDetail.tsx:269-280`) can't show who launched the run.
- **Misleading "Scoring profile: auto"**: per `docs/rfc/EVAL-CUSTOM-METRIC-RFC.md`, profile auto-detection no longer drives DeepEval metric selection (replaced by an explicit `metrics` list, already read back in `EvaluationRun.metrics`) — `profile` now only affects structural checks (`resolve_profile()` in `workers/activities.py:152`). Displaying "Scoring profile: auto" alongside "Judge" implies it still governs scoring globally, which it doesn't anymore.

**Fix**:
- Add `created_by` to the `EvaluationRun` schema and surface it as "Author" in the run metadata panel.
- Rename or document the "Scoring profile" field to reflect its real (now narrower) role, or drop it from the summary panel if it's not meaningful to the end user.

### 6. Team settings "← Back" always lands on Agents, regardless of context

`TeamContentNavbar.tsx:229-244` — the "← Back" button shown above every settings tab (Members, Settings, Activity, Evaluations, Model routing) and above Usage is **not** a history-based back (`navigate(-1)`); it's hardcoded to `navigate(\`/team/${teamId}/agents\`)`, duplicated identically in the `inSettings` and `inUsage` branches. No matter which settings tab you're on or which page you opened settings from, "Back" always drops you on the Agents page.

**Fix**: make "Back" return to wherever the user actually came from (history-based), or at minimum to a consistent, sensible destination (e.g. team overview) instead of a hardcoded route. Audit whether any other "Back"/breadcrumb links in the settings panel have the same issue.

---

## Scope

Spans both repos:
- `fred` (this repo) — frontend fixes for #1, #2, #3, #5 (frontend part), #6.
- `fred-agent-evaluator` — backend fixes for #3 (list endpoints), #4, #5 (backend schema).

## Out of scope / follow-ups

- Splitting into separate issues per repo/point if the fixes don't land together — left to whoever picks this up.

CohenOdelia:
# PR — Evaluation UI/backend polish: 6 fixes from a walkthrough

**Issue:** #2141 — Evaluation UI/backend polish: 6 fixes from a walkthrough

**Problem in one sentence:** The evaluation settings UI had several product gaps and misleading behaviors — redundant rerun controls, no typed/versioned JSON export, no list scaling on evaluations/runs, missing run-author metadata, misleading scoring-profile wording, and a hardcoded "Back" that always dropped users on Agents.

**Backlog ref:** n/a — tracked directly in #2141. This is the frontend slice; the backend fixes (#3 list endpoints, #4 doc, #5 schema) live in the fred-agent-evaluator repo on branch `45-…` (separate PR).

- **RFC written?** Not required — frontend/product polish against an already-tracked issue. No new backend contract or design surface is introduced here; the UI consumes existing/generated API fields (`created_by`, paginated list params) that were added in the backend PR.
- **Plan presented?** Yes — the six fixes were scoped from the walkthrough issue and implemented incrementally with per-fix checkpoints.
- **Confirmation received?** Yes — implementation was explicitly requested and refined live during the session.

## 3. What you built

Frontend fixes for #2141 in `apps/frontend/src/rework/.../TeamSettingsEvaluations/` and `TeamContentNavbar.tsx`:

- **#1** — Removed the redundant header "Rerun" (most-recent) button; kept "+ New run". Per-row rerun stays. Dropped the now-dead `mostRecentRun`/`rerunManagedTarget` derivations.
- **#2** — Typed/versioned JSON export (`evaluationExport.ts`): case-level (`{schema_version, kind, exported_at, case}` download) next to the existing clipboard copy, and run-level (downloads the backend's already-versioned `RunReportResponse` — all cases + metric averages + analysis when present).
- **#3** — Search-by-name + sort + real pagination on the evaluations list; explicit sort + pagination on the runs list (both wired to the new offset/limit/sort/q + total backend contract).
- **#5 (frontend)** — Surfaced the run author (`created_by`) in the Run information panel; relabelled the misleading "Scoring profile" as the structural-check profile.
- **#6** — Settings/Usage "Back" is now history-based (`navigate(-1)`) with a fallback to the team's agents view when there's no history; both hardcoded branches fixed.
- Regenerated the evaluation OpenAPI client (`evaluationOpenApi.ts`) to pick up `created_by` and the paginated list contract.
- (The commit also carries agent-name labelling for run targets — `buildAgentLabels`/`managedInstances` — added on top of the above.)

## 4. Proof of quality

**Tests added or updated:**

|------|------|-----------------|
| none added | n/a | Frontend UI wiring + generated-client refresh. The existing suite (692 tests) still passes; see gap note below. |

**Honest gap:** the pagination offset/reset logic and `evaluationExport.ts` (envelope shape, filename) are unit-testable and currently have no dedicated test. Worth a follow-up.

**`make code-quality` (last lines):**

```
npx tsc --noEmit
npx prettier . --check
All matched files use Prettier code style!
npx eslint "src/**/*.{ts,tsx}"
All frontend code quality checks completed        (exit 0)
```

Note: 3 files in the original commit were unformatted — fixed with `prettier --write` (formatting-only, now uncommitted pending amend).

`basedpyright`: n/a — frontend-only PR, no Python package in this repo touched. (The backend baseline lives in the fred-agent-evaluator PR.)

**`make test` (summary):**

```
Test Files  66 passed (66)
     Tests  692 passed (692)
  Duration  20.52s
```

## 5. Docs updated

|---|---|
| Backlog `[ ]` item now done | n/a here — `EVAL-LLM-ANALYSIS-BACKLOG.md` (#4) is updated in the fred-agent-evaluator PR, not this repo |
| New behaviour, API field, or contract change | `evaluationOpenApi.ts` regenerated (generated client — the contract itself changed in the backend PR) |
| Frozen contract touched (`execution.py`, `agent_app.py`, OpenAPI) | n/a — no backend contract file in this repo touched |
| UX component implemented or visual status changed | not updated — `docs/swift/ux/COMPONENT-UX.md` was not touched; these are polish edits to existing views, no new component. Flag if a status line is expected. |
| Phase progress row | n/a |
| WORKPLAN sprint item finished | n/a — frozen doc, tracking is GitHub #2141 |

## 6. Close-out statement

```
## Task close-out
- Code: frontend fixes for #2141 (#1 header rerun removed, #2 typed/versioned JSON export, #3 search/sort/pagination on both lists, #5 run author + relabelled profile, #6 history-based Back) + regenerated evaluation client; prettier applied to 3 unformatted files.
- Tests: none added (UI wiring + client refresh); existing suite green — 692 passed / 66 files. Gap: pagination + evaluationExport untested (follow-up).
- Docs updated: evaluationOpenApi.ts (generated). COMPONENT-UX.md not updated (no new component). Backend docs (#4) live in the fred-agent-evaluator PR.
- Backlog: none in this repo — tracked on GitHub #2141.
- Skipped steps: Step 1 RFC (not required — product polish, no new contract); Step 3.5 issue already exists (#2141).
```

## 7. Risk and rollback

**What breaks if this is wrong?** Purely the evaluation settings UI (list scaling, export buttons, run metadata, Back nav). No backend behavior, no shared component outside these views except `TeamContentNavbar` Back — worst case a wrong Back destination, non-destructive. Requires the backend PR (`45-…`) to be deployed for pagination/`created_by` to return data; against the old backend the lists still render (params ignored, author blank).

**How do we roll back?** Revert the frontend commit(s) on this branch — no migrations, no persisted state, no contract owned here.

### DIFF
--- a/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
+++ b/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
@@ -121,30 +121,6 @@
   text-align: left;
 }
 
-.nameHeaderCell {
-  transform: translateX(-4px);
-}
-
-.targetHeaderCell {
-  transform: translateX(-38px);
-}
-
-.stateHeaderCell {
-  transform: translateX(-68px);
-}
-
-.verdictHeaderCell {
-  transform: translateX(-104px);
-}
-
-.progressHeaderCell {
-  transform: translateX(-140px);
-}
-
-.scoresHeaderCell {
-  transform: translateX(-176px);
-}
-
 .bodyRow {
   position: relative;
   cursor: pointer;
--- a/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
+++ b/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
@@ -60,6 +60,17 @@
   flex: 1 1 180px;
 }
 
+.toolbar {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: var(--spacing-m);
+}
+
+.toolbarSpacer {
+  flex: 1 1 auto;
+}
+
 /* ── Grid table ──────────────────────────────────────────────────────────── */
 
 .table {
@@ -73,10 +84,21 @@
 
 .row {
   display: grid;
-  grid-template-columns: 1.6fr 1fr auto auto 160px 180px auto;
+  grid-template-columns:
+    minmax(0, 0.78fr)
+    minmax(0, 0.9fr)
+    minmax(0, 0.62fr)
+    minmax(0, 0.72fr)
+    minmax(0, 0.9fr)
+    minmax(0, 0.95fr)
+    max-content;
   align-items: center;
-  gap: var(--spacing-m);
-  padding: var(--spacing-s) var(--spacing-m);
+  column-gap: var(--spacing-s);
+  padding: var(--spacing-s) var(--spacing-xs);
+}
+
+.row > * {
+  min-width: 0;
 }
 
 .headerRow {
@@ -85,14 +107,53 @@
   font: var(--font-label-small-emphasized);
   letter-spacing: 0.04em;
   text-transform: uppercase;
+  padding-left: var(--spacing-xs);
+  padding-right: var(--spacing-xs);
+}
+
+.nameHeaderCell,
+.targetHeaderCell,
+.stateHeaderCell,
+.verdictHeaderCell,
+.progressHeaderCell,
+.scoresHeaderCell,
+.actionsHeaderCell {
+  text-align: left;
+}
+
+.nameHeaderCell {
+  transform: translateX(-4px);
+}
+
+.targetHeaderCell {
+  transform: translateX(-38px);
+}
+
+.stateHeaderCell {
+  transform: translateX(-68px);
+}
+
+.verdictHeaderCell {
+  transform: translateX(-104px);
+}
+
+.progressHeaderCell {
+  transform: translateX(-140px);
+}
+
+.scoresHeaderCell {
+  transform: translateX(-176px);
 }
 
 .bodyRow {
   position: relative;
   cursor: pointer;
-  border-top: 1px solid var(--outline-variant);
   color: var(--on-surface);
   font: var(--font-body-medium);
+}
+
+.bodyRow + .bodyRow {
+  border-top: 1px solid var(--outline-variant);
 }
 
 .bodyRow:hover {
@@ -130,6 +191,18 @@
   gap: var(--spacing-xs);
 }
 
+.targetCell {
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-2xs);
+  min-width: 0;
+}
+
+.targetName {
+  color: var(--on-surface);
+  font: var(--font-body-medium-emphasized);
+}
+
 .name {
   color: var(--on-surface);
   font: var(--font-body-medium-emphasized);
@@ -146,11 +219,17 @@
   display: flex;
   flex-direction: column;
   gap: var(--spacing-2xs);
-  /* Grid items default to min-width: auto, letting long content (e.g. a
-     long judge_profile_id like "MISTRAL-SMALL") blow past this column's
-     180px track and spill into the actions column next to it. Allow the
-     item to shrink to its track so the pill below can ellipsis instead. */
   min-width: 0;
+}
+
+.stateCell,
+.verdictCell {
+  display: flex;
+  align-items: center;
+}
+
+.stateCell {
+  padding-left: var(--spacing-m);
 }
 
 .scoreLine {
@@ -172,13 +251,49 @@
 
 .actionsCell {
   display: flex;
-  flex-shrink: 0;
   justify-content: flex-end;
-  gap: var(--spacing-xs);
-  /* Grid equivalent of flex-shrink: 0 — never let this column collapse
-     below the Detail/Delete buttons' natural width, whatever the sibling
-     scoresCell column does. */
+  gap: var(--spacing-s);
   min-width: max-content;
+}
+
+.actionsHeaderCell {
+  left: 0;
+  justify-self: end;
+}
+
+.nameHeaderCell,
+.nameCell {
+  grid-column: 1;
+}
+
+.targetHeaderCell,
+.targetCell {
+  grid-column: 2;
+}
+
+.stateHeaderCell,
+.stateCell {
+  grid-column: 3;
+}
+
+.verdictHeaderCell,
+.verdictCell {
+  grid-column: 4;
+}
+
+.progressHeaderCell,
+.progressCell {
+  grid-column: 5;
+}
+
+.scoresHeaderCell,
+.scoresCell {
+  grid-column: 6;
+}
+
+.actionsHeaderCell,
+.actionsCell {
+  grid-column: 7;
 }
 
 .muted {
@@ -277,5 +392,14 @@
 .caseLink {
   align-self: flex-end;
   color: var(--primary);
+  font: var(--font-label-medium);
+}
+
+.pagination {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  gap: var(--spacing-s);
+  color: var(--on-surface-retreat);
   font: var(--font-label-medium);
 }
--- a/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
+++ b/apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
@@ -92,8 +92,8 @@
     minmax(0, 0.9fr)
     minmax(0, 0.95fr)
     max-content;
-  align-items: center;
   column-gap: var(--spacing-s);
+  align-items: center;
   padding: var(--spacing-s) var(--spacing-xs);
 }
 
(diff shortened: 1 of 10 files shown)


apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.tsx
```diff
@@ -187,7 +214,12 @@ export default function EvaluationRuns({
     }
   };
 
-  const runs = data ?? [];
+  // The list endpoint is now paginated: it returns { runs, total } rather than a
+  // bare array. `total` is the full server-side count (not this page's length).
+  const runs = data?.runs ?? [];
```
dimitri-tombroff, line 242: [P2] Keep the KPI cards global after pagination

`runs` now contains only the current page, but the four KPI values below (`running`, `completed`, `totalCases`, and `criticalErrors`) are still derived from it. For example, with oldest-first sorting the UI can show 0 active runs while active runs exist on a later page. Please consume global aggregates from the backend (or a separate summary query) rather than deriving dashboard totals from this page slice.

apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.tsx
```diff
@@ -408,6 +432,30 @@ export default function EvaluationRuns({
         </div>
       )}
 
+      {!isLoading && total > RUNS_PAGE_SIZE && (
```
dimitri-tombroff, line 454: [P2] Return to a valid page after deleting its last row

If there are 26 runs, the user opens page 2 (`offset = 25`), and deletes its only row, `refetch()` returns `{ runs: [], total: 25 }` while the offset remains 25. This condition then hides the pager, leaving the user on an empty page until a reload or sort change. Please decrement/reset the offset after deleting the last row of a non-first page, or clamp it whenever `currentPage > pageCount`.

apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
```diff
@@ -73,28 +84,78 @@
...
+.verdictHeaderCell {
+  transform: translateX(-104px);
+}
+
+.progressHeaderCell {
+  transform: translateX(-140px);
+}
+
+.scoresHeaderCell {
+  transform: translateX(-176px);
```
dimitri-tombroff: [P2] Keep headers aligned by the shared grid

The header and body already use the same grid tracks, so these cumulative fixed translations (`-4px` through `-176px`) move each label away from the data column it describes and become increasingly incorrect as the available width changes. Please remove the per-column transforms and let the grid align headers and cells; if a small inset is needed, apply the same tokenized padding to the corresponding header and body cell.

dimitri-tombroff (commented):
Assuming the matching `fred-agent-evaluator` backend contract is deployed, I found no P0/P1 in the FRED-only part. I am leaving three P2 frontend comments below.

apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.tsx
dimitri-tombroff: Fixed in d3e1b77c: the offset now snaps back to the last valid page whenever `total` drops below it (delete-triggered or otherwise), so a now-empty page never hides the pager or shows a misleading empty state.

apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.module.css
dimitri-tombroff: Fixed in d3e1b77c: removed the six per-column `translateX` rules — header and body already share the same `.row` grid template and `grid-column` mapping, so they align without any manual offset.

apps/frontend/src/rework/components/shared/organisms/TeamSettingsPanel/TeamSettingsEvaluations/views/EvaluationRuns.tsx
dimitri-tombroff: Fixed in 6adc5eab: the KPI cards now consume a new evaluation-wide aggregate — GET /evaluations/{id}/runs/summary, added in fred-agent-evaluator@2e287ff (fred-agent/fred-agent-evaluator#46) — instead of deriving from the current page of the runs list. This depends on PR #46 (still open, not yet merged); until it's deployed the summary endpoint 404s and the cards show 0.

dimitri-tombroff (approved):

PR #2236, opened 2026-08-05
RFC: Capability scope as an authorization ceiling (CAPAB-SCOPE-01)

Maxime-DARAGON:
## What this is

An RFC (design doc only — **no code**) proposing **agent-scope-first authorization**: capability selection derives a hard authorization *ceiling* that is enforced **before** user ReBAC, uniformly across every document-touching port.

`docs/swift/rfc/CAPABILITY-SCOPE-CEILING-RFC.md`

## Why I need your read before we commit to it

This is a decision about an authorization *principle* that will apply to every future capability, so I want your sign-off on the direction before opening any implementation issue. The core claim:

> Effective access = **agent ceiling ∩ user rights**, ceiling evaluated first. An "attachments-only" agent must never surface a corpus document, **even for a user who has full corpus access.**

Today only `document_access.search` implements this (as the per-tool `attachments_only` boolean → structural corpus-branch skip in KF). Every other document tool (`summarize`, and by construction any future PPT/tabular/export capability) authorizes **user-rights-first, with no agent ceiling** — `DocumentSummarizeAdapter` states it outright ("KF's own per-document ReBAC is the real gate"). The RFC argues we should generalize search's model into a once-declared ceiling that all ports honor, rather than N per-tool booleans that fail open by omission.

## Points I'd specifically like your view on

- **§2 / §5** — this deliberately stays inside RUNTIME-07 rev. 2 / D5 (no control-plane-signed grant; pod stays the execution authority). Tier 1 = pod-enforced ceiling (no wire change); Tier 2 = KF-boundary enforcement, which I've framed to **converge with your `AGENT-FILESYSTEM-HARDENING` P1 actor principal** rather than invent a second mechanism. Does that convergence hold up, or do you see them as genuinely different principals?
- **§4** — the risks-if-we-don't (confused-deputy corpus exposure; ceiling eroding silently per new capability; the "restricted agent / privileged user" case being unbuildable today).
- **§8** — four open questions I've left for decision (scope granularity; fate of `document_access.search_attachments_only`; Tier-1-only-first; sequencing vs. the standalone attachment/summarize capability discussion).

No implementation is approved by this RFC alone — I'll only open a scoped issue once the principle is agreed.

### DIFF
--- a/docs/swift/rfc/CAPABILITY-SCOPE-CEILING-RFC.md
+++ b/docs/swift/rfc/CAPABILITY-SCOPE-CEILING-RFC.md
@@ -1,0 +1,273 @@
+# RFC: Capability scope as an authorization ceiling (agent-scope-first authorization)
+
+**Status:** proposed — open design question; no implementation approved by this RFC alone
+**Author:** Maxime Daragon
+**Date:** 2026-08-05
+**ID:** CAPAB-SCOPE-01
+**Related docs:** `docs/swift/design/RUNTIME-EXECUTION-CONTRACT.md` §8.11 (pod-side authorization), §8.15 amendment 2026-07-21 (`attachments_only`), §8.16 (document ports); `docs/swift/rfc/AGENT-FILESYSTEM-HARDENING-RFC.md` (F1/P1 — same problem at the `/fs` boundary); `docs/swift/capabilities/AUTHORING.md`; `docs/swift/platform/REBAC.md`
+
+---
+
+## 1. Problem
+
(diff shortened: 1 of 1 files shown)


Maxime-DARAGON:
Cross-link: the frontend that presents capabilities as independent packs just landed in #2250 (front-only, no backend change). It's the presentation layer on top of the model this RFC decides — its decision gates the real backend cleanup (authorization ceiling + whether "packs" become a first-class server-side concept). See the detailed note on #2250.

PR #2241, opened 2026-08-05
feat(frontend): session-history cache for reducing conversation switches latencies (#2239)

### LINKED ISSUE #2239: Chat: cache opened conversations client-side so switching back doesn't reload everything
## Problem

Switching between conversations in `ManagedChatPage` reloads everything, every time — including when returning to a conversation opened seconds earlier:

- The sessionId-change effect (`useManagedChat.ts`) calls `reset()`, which blanks the whole thread synchronously.
- `useSessionHistory` then round-trips `prepare-execution` (POST) + a raw history `fetch` before anything renders; a second `prepare-execution` fires in parallel from `prepareChatControls`. Nothing is cached: `loadedRef` only remembers the single last-loaded session, so A → B → A always refetches A.
- While loading, `isLoadingHistory` disables the composer and shows a spinner — the UI goes blank-then-spinner on every switch.

Two latent races live in the same code path:

- `useSessionHistory` has no stale-response guard: switching A → B while A's history fetch is in flight lets A's response land via `replaceAllMessages` under B's session.
- A history response arriving while a turn is streaming would clobber the live turn (today only the `msgs.length > 0` guard incidentally protects the brand-new-session case).

## Proposal (client-side only, per the History Ownership Contract)

`CHAT-UI-BACKLOG.md` §0.4 forbids the control-plane from caching/serving message history, so the cache must live in the frontend:

- Add a module-level, bounded (LRU, ~20 sessions) in-memory cache of `ChatMessage[]` keyed by `session_id`, surviving the per-agent `key={agentInstanceId}` remount.
- On switch to a cached session: render the cached thread instantly (no spinner, composer stays enabled), then revalidate in the background and swap in the fresh history when it arrives (stale-while-revalidate).
- On leaving a session: snapshot the currently displayed messages into the cache so live-streamed turns are included without an extra fetch.
- Guards: a history response is applied (and cached) only if its session is still the active one and no turn is in flight — fixing both races above.

Out of scope: deduplicating the double `prepare-execution` POST per switch (worth a separate look), any control-plane/runtime change.

## Acceptance

- A → B → A: A's messages render instantly from cache; a background refresh reconciles with the runtime.
- No stale cross-session rendering when switching during an in-flight history load.
- An in-flight streamed turn is never replaced by a late history response.
- Cache is bounded; unit tests cover hit, miss, stale-response, in-flight-turn, and eviction paths.

simcariou:
Closes #2239.

## What

Switching between conversations in `ManagedChatPage` no longer reloads everything. Returning to a previously opened conversation renders the thread instantly — no blank flash, no spinner, composer stays enabled — while the runtime history is silently revalidated in the background and swapped in when it arrives.

## How

- **`sessionHistoryCache.ts` (new)** — a module-level, bounded LRU (20 sessions) mapping `session_id` → `ChatMessage[]`. Module-level on purpose: it survives the `key={agentInstanceId}` remount when hopping between two agents' conversations. Client-side only, per the History Ownership Contract (`CHAT-UI-BACKLOG.md` §0.4) — the control-plane still never proxies, caches, or serves message history, and the runtime stays the source of truth.
- **`useSessionHistory`** — serve-then-revalidate: on a cache hit the thread is applied synchronously and `isLoading` never flips (the fetch becomes a silent background revalidation); on a miss, behavior is unchanged. The async path gains explicit guards — a response is applied and cached only if
  1. its session is still the active one (fixes a latent race: switching A → B while A's fetch was in flight could render A's history under B),
  2. no turn is streaming (a late history response can no longer clobber a live turn),
  3. it is non-empty (preserves the existing protection of a brand-new session's optimistic first message).
- **`useManagedChat`** — on leaving a session (switch or unmount), the displayed thread is snapshotted into the cache via an effect cleanup that runs before the next session's reset wipes the state. This folds live-streamed turns into the cache with zero extra fetches. Snapshot is skipped while the initial history load is still in flight, so a half-loaded thread is never cached as complete.

## Tests

- New `useSessionHistory.test.tsx` (8 tests): cache miss, cache hit + background revalidation, stale-response-after-switch, empty-response, live-turn guard, no re-fetch on effect re-fire, LRU eviction, LRU recency refresh.
- `useManagedChat.test.tsx`: new departure-snapshot wiring test; existing 23 tests untouched and green.
- Full frontend suite: 105 files / 1012 tests passed; `make code-quality` (tsc + prettier + eslint) clean.

## Docs

- `docs/swift/backlog/CHAT-UI-BACKLOG.md` §1.4 — dated entry documenting the cache and its guards; §0.4 explicitly reaffirmed.

## Out of scope

The double `prepare-execution` POST per switch (one from the reset effect's `prepareChatControls`, one from `useSessionHistory`) is unchanged — worth a separate consolidation pass.

### DIFF
--- a/apps/frontend/src/rework/components/pages/ManagedChatPage/sessionHistoryCache.ts
+++ b/apps/frontend/src/rework/components/pages/ManagedChatPage/sessionHistoryCache.ts
@@ -14,32 +14,35 @@
 
 import type { ChatMessage } from "../../../../slices/agentic/agenticOpenApi";
 
-// In-memory per-session history cache backing instant conversation switches
-// (#2239). Module-level ON PURPOSE: ManagedChatPage remounts with
-// key={agentInstanceId} when switching agents, so hook/ref state would be
-// lost exactly when the user hops between two agents' conversations — the
-// case the cache exists for. Client-side only, per the History Ownership
-// Contract (CHAT-UI-BACKLOG.md §0.4): the control-plane must never cache or
-// serve message history, and the runtime stays the source of truth — every
-// cached entry is revalidated against it on re-entry (useSessionHistory's
-// serve-then-revalidate).
+// Per-session history cache backing instant conversation switches (#2239).
+//
+// Two layers, same bounded LRU semantics:
+// - a module-level Map (the authority while the page lives) — module-level ON
+//   PURPOSE: ManagedChatPage remounts with key={agentInstanceId} when
+//   switching agents, so hook/ref state would be lost exactly when the user
+//   hops between two agents' conversations;
+// - a best-effort sessionStorage mirror, so the cache also survives a page
+//   refresh (F5). sessionStorage — not localStorage — deliberately: it is
+//   per-tab and cleared when the tab closes, the same lifetime already
+//   accepted for `chat.composer.<sessionId>`, so conversation content never
+//   outlives the user's tab on a shared machine. Every storage access is
+//   wrapped: quota exhaustion or a blocked storage API degrades to
+//   in-memory-only, never to an error.
+//
+// Client-side only, per the History Ownership Contract (CHAT-UI-BACKLOG.md
+// §0.4): the control-plane must never cache or serve message history, and the
+// runtime stays the source of truth — every cached entry is revalidated
+// against it on re-entry (useSessionHistory's serve-then-revalidate).
 const MAX_CACHED_SESSIONS = 20;
+const STORAGE_PREFIX = "chat.history.";
+const STORAGE_INDEX_KEY = "chat.history#index";
 
 // Map preserves insertion order; delete-then-set on every read and write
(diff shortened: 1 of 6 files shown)


simcariou:
Added per-tab persistence: the LRU is now mirrored into `sessionStorage` (`chat.history.<sid>` + `chat.history#index`), so the conversation also renders instantly after a page refresh (F5) instead of blanking behind a spinner, then revalidates as usual. `sessionStorage` on purpose — per-tab, cleared on tab close, same lifetime as `chat.composer.*` — so no conversation content persists on disk beyond the tab (GDPR posture unchanged vs `localStorage`). Quota-safe: storage failures degrade to in-memory-only. +4 tests (reload survival, cross-reload eviction, quota degrade, corrupted-entry tolerance); suite at 1016 green.

Issue #2274, opened 2026-08-07
[P1] Markdown preview OOMs the browser too: unbounded extraction, per-byte base64 decode, whole-tree render


simcariou:
Sibling of #2273 (native PDF rendering). That issue bounded the `.pdf` "Fichier"
path; this one is the *other* render strategy, and it is unbounded end to end.

## Why this still matters after #2273

`DocumentViewer` picks between two strategies (`DocumentViewer.tsx:57-79`). The
markdown extraction path serves:

- **every non-PDF format** — `.docx`, `.xlsx`, `.csv`, `.pptx`, … all render
  through `MarkdownDocumentBody`, never through a native renderer;
- **the "Raw" tab of the PDF viewer itself**.

So the 50 MB PDF that no longer crashes the tab in "Fichier" mode can still
crash it one click away in "Raw", and a 50 MB `.docx` was never covered at all.

### 1. The base64 decode is a per-byte JS closure over the whole payload

`decodeMaybeBase64Utf8` (`documentViewerUtils.ts:22-30`):

```ts
const binary = atob(value);
const bytes = Uint8Array.from(binary, (char) => char.charCodeAt(0));
return new TextDecoder("utf-8", { fatal: true }).decode(bytes);
```

`Uint8Array.from` with a mapper invokes that closure **once per byte** — for a
30 MB extraction, ~30 million synchronous calls on the main thread. Worse, at
peak the payload is resident four times over: the base64 string, the binary
string, the `Uint8Array`, and the decoded UTF-16 result.

### 2. Nothing bounds what goes into the renderer

`MarkdownRenderer` (`MarkdownRenderer.tsx:169-226`) hands the entire string to
`ReactMarkdown` with five remark plugins and three rehype plugins. That means, on
the full document: an mdast parse, a hast conversion, a **complete tree rebuild**
by `rehypeCitations` (`:84-130`, allocating fresh nodes as it walks), a
`rehype-katex` pass, a `rehype-sanitize` pass — and then React renders *every
resulting node* into the DOM. No truncation, no windowing, no cap on `text`.

### 3. The payload is held twice on the client

`MarkdownDocumentBody` copies the decoded string into component state
(`DocumentViewer.tsx:118,141`) while RTK Query independently caches the response
body — so the document sits in memory at least twice before rendering starts.

### 4. The endpoint returns the whole extraction in one JSON body

`GET /markdown/{document_uid}` → `MarkdownContentResponse { content: str }`
[...]

PR #2278, opened 2026-08-07
fix(FRONT): PDF preview no longer OOMs the tab — virtualize pages and fetch by byte range

### LINKED ISSUE #2273: [P1] PDF preview OOMs the browser: the viewer mounts a canvas for every page of the document
Split out of #2272 (upload half stays there).

## Symptom

Opening a ~50 MB PDF in the document viewer makes the browser tab run out of
memory and crash. The UI advertises up to 200 MB per file
(`translation.json:546` — *"max 200Mo par fichier"*), so this is well inside the
supported range.

## Confirmed cause

`PdfStreamingDocumentViewer` built its page list as
`Array.from({ length: numPages ?? 0 }, ...)` and handed **all of them** to
`<Document>` at once. No virtualization, no page window: every single page of
the document was mounted as a live `<canvas>` at `pageWidth` (≈800 px), each
with its own annotation layer.

A rendered page canvas at ~800×1100 costs roughly 3.5 MB of bitmap. A 50 MB PDF
is typically several hundred to a few thousand pages, so the viewer tried to
allocate on the order of gigabytes the instant `onLoadSuccess` fired — hence a
hard tab crash rather than a slow render.

Contributing factor: no pdf.js transport options were passed to `<Document>`, so
`disableAutoFetch` was off and pdf.js eagerly pulled the whole file.

**Not the server's fault**: `/raw_content/stream/{document_uid}` already
advertises `Accept-Ranges: bytes` and implements 206/416 correctly
(`content_controller.py:300-369`). The client simply never asked for ranges.

## Fix

- **B1 — virtualize the page list.** Render one cheap, correctly-sized
  placeholder slot per page (so the scrollbar stays honest) and mount a real
  `<Page>` canvas only for slots inside a band around the viewport, via a single
  `IntersectionObserver`. Placeholders are sized from page 1's real geometry,
  with A4 portrait as the fallback.
- **B2 — on-demand fetching.** Pass a module-level (stable-identity)
  `options={{ disableAutoFetch: true, disableStream: false }}` to `<Document>`
  so pdf.js uses the byte-range support the backend already provides instead of
  buffering the whole file up front.
- **B3 — guard rail.** Past 500 pages, show an opt-in panel ("Document
  volumineux — afficher quand même") instead of auto-rendering: virtualization
  bounds the canvases, but one placeholder plus one observer entry per page is
  not free and pdf.js still walks the whole page tree.

## Acceptance criteria

- A 50 MB / several-hundred-page PDF opens, scrolls, and never holds more than a
  bounded number of rendered canvases.
- The scrollbar reflects the document's real length while pages are unrendered.
- No proxy in front of knowledge-flow strips or buffers `Range` requests
  (no nginx config is checked into this repo — needs a check against the
  deployed chart).

simcariou:
Closes #2273.

Opening a ~50 MB PDF crashed the browser tab, well inside the 200 MB per file the UI advertises.

## What was wrong

Two independent problems, both of which had to be fixed to bound memory:

1. **Every page was mounted as a live `<canvas>`.** The page list was built with `Array.from({ length: numPages })` and handed to `<Document>` whole. At roughly 3.5 MB of bitmap per page, a several-hundred-page document tried to allocate gigabytes on `onLoadSuccess`.
2. **The whole file was downloaded and held in memory**, regardless of what was displayed.

## What changed

- **Virtualization** — one cheap, correctly-sized placeholder per page keeps the scrollbar honest; a real `<Page>` is mounted only inside a 600 px band around the viewport, tracked by a single `IntersectionObserver`. Slots are sized from page 1's own geometry (A4 portrait as fallback) so landscape and slide-shaped documents scroll correctly too.
- **Byte-range fetching** — `options={{ disableAutoFetch: true, disableStream: true }}`, module-level so the object identity stays stable (a fresh object per render makes react-pdf reload the document).
- **Guard rail** — past 500 pages, an opt-in panel instead of auto-rendering. Virtualization bounds the canvases, but one placeholder plus one observer entry per page is not free and pdf.js still walks the whole page tree.
- The component's hardcoded strings move to i18n (fr/en) alongside the new copy.

### `disableStream` is the load-bearing flag

`disableAutoFetch` alone is not enough, and this cost a round trip to find out. It only suppresses the background prefetch of remaining chunks once ranges are in use — it never touches the initial full-file request. pdf.js cancels that one only when streaming is off:

```js
if (!this._isStreamingSupported && this._isRangeSupported) {
  this.cancel(new AbortException("Streaming is disabled."));
}
```

Left at `disableStream: false`, the first version of this fix still pulled the entire document: one `200` of 51.86 MB followed by 36 partial `206`s — 54 MB on the wire for a 52 MB file, fully resident in memory on top of the rendered pages. The virtualization was working; the fetching was not.

## Verification

Measured in the browser on a 51.86 MB document:

| | Requests | Transferred |
|---|---|---|
| Before | 1 × `200` (51.86 MB) + 36 × `206` | 54.17 MB |
| After | 1 × aborted discovery (~82 kB) + `206` per page on scroll | bounded by what is displayed |

The single `NS_BINDING_ABORTED` entry is expected and is the signature of the fix: pdf.js must open a full request to read `Accept-Ranges` / `Content-Length` from the response headers, then aborts it as soon as they confirm range support.

Also verified that no proxy on the path strips `Range`: both the Vite dev proxy and the `/knowledge-flow/` block of the nginx shipped in the frontend image forward `Range` verbatim and return `206` / `Content-Range` / `Accept-Ranges` intact.

## Deployment note

**Nothing needs to be enabled for ranges** — nginx and Envoy forward `Range` natively, no annotation required. But one setting silently breaks this fix: **compressing `application/pdf`** anywhere on the path (ingress-nginx `gzip-types`, a CDN, a WAF) strips both `Accept-Ranges` and `Content-Length` from the initial response, and pdf.js falls back to buffering the whole file — reintroducing the OOM with no other visible symptom. Verified:

| Initial response | `Accept-Ranges` | `Content-Length` |
|---|---|---|
| With `gzip_types application/pdf` | **absent** | **absent** (chunked) |
| Without compression | `bytes` | present |

Worth knowing that the chart's default Ingress routes `/knowledge-flow` **directly** to `knowledge-flow-backend:8000`, bypassing the frontend pod's nginx — so the proxy to audit is the ingress controller, not the frontend container.

## Tests

11 tests in `PdfStreamingDocumentViewer.test.tsx`, covering the virtualization window, placeholder geometry, the 500-page guard, and the transport options. `make code-quality` green.

## Not in scope

Sibling issue #2274: the markdown extraction path (`MarkdownDocumentBody` → `MarkdownRenderer`) is unbounded end to end and serves this viewer's own "Raw" tab, so a 50 MB PDF that no longer crashes in "Fichier" mode can still crash one click away.

Also documents a pre-existing code/doc divergence fix in `COMPONENT-UX.md`: pages render at `PDF_SCALE = 1.0`, not the documented "fixed 0.8 scale".


simcariou:
Third commit (`75abeb9fd`) — the guard now fires **before** parsing, after field testing surfaced a case the first two commits didn't cover.

**Symptom**: a 3666-page text-only PDF still read its entire 50 MB at open (803 serial 64 KB range requests, ~48 s), while the 36-page image PDF of the same size streamed on scroll as intended. The page-count guard never helped: it renders from `onLoadSuccess`, i.e. after the parse that *is* the expensive act.

**Root cause** (verified by probing pdf.js in Node against a range-logging server): at open, `checkLastPage` validates `/Count` by fetching the **last** page's dict, and `Catalog.getPageDict` walks `/Kids` linearly, fetching every kid just to learn it's a one-page leaf. With a **flat** 3666-ref `/Kids` array — legal, and what naive generators emit — that fetches all 3666 page dicts, interleaved with their content streams across the whole file. Not a text-vs-images difference: a page-count × tree-shape difference.

Control experiment: the same 3666 pages in a fanout-16 **balanced** tree (what LaTeX/Word/Acrobat emit — intermediate `/Count` nodes let the walk skip subtrees unfetched):

| Structure | Open cost (no page displayed) |
|---|---|
| Flat `/Kids`, 50 MB | 52 × 1 MB requests — the whole file |
| Balanced, **91 MB** | **2 requests, 1.3 MB** |

**Change**: the viewer probes the file's size with a one-byte ranged GET (`Content-Range` carries the total) and, past 20 MB, shows the opt-in panel **without mounting `<Document>`** — pdf.js does no work until the user asks. Size is the only property knowable before parsing, so it's the only pre-parse signal available; a failed probe falls through to the unguarded path. Plus `rangeChunkSize: 1 MB`, which cuts the pathological case's serial round trips 803 → 54 and costs nothing on well-formed files.

17 tests now cover the suite, including the probe contract (one byte, fall-through on failure or mangled `Content-Range`).

Issue #1460, opened 2026-04-02
"Too many inputs in request, split into more batches" in medium/rich pdf ingestion pipeline


simcariou:
```
{

  "type": "workflowExecutionFailedEventAttributes",

  "failure": {

    "message": "Child Workflow execution failed",

    "cause": {

      "message": "Child Workflow execution failed",

      "cause": {

        "message": "Activity task failed",

        "cause": {

          "message": "vectorization processing failed",

[...]

PR #2285, opened 2026-08-07
fix(knowledge-flow): remember learned embedding batch cap across bulk slices

### LINKED ISSUE #2282: Vector ingestion re-pays Mistral 400s on every bulk slice — adaptive embedding batching never remembers the learned batch size
## Problem

When ingesting a large document (~40k chunks observed), every `bulk_size`=1000 slice is first sent as a single Mistral `/v1/embeddings` request. Mistral rejects it (HTTP 400, code 3210 `invalid_request_prompt` — the limit is on **total tokens per request**), the adaptive splitter then retries 1000 → 500 → 250, which succeeds.

The splitter works (nothing is lost — it was introduced by #1460), but **nothing is remembered**:
- the next slice starts again at 1000 and re-pays 2–3 failed provider calls (~10–15 s each cycle);
- within a slice, the right half of a binary split re-attempts a size the left half just saw fail.

On the observed document this wastes on the order of **~100 failed Mistral calls and several minutes of pure latency** per ingestion, plus API quota and log noise.

## Fix (agreed with dev)

Make the adaptive batching *sticky* in `OpenSearchVectorStore` (`apps/knowledge-flow-backend/.../core/stores/vector/opensearch_vector_store.py`):
- remember, on the store instance, the largest batch size the provider has accepted after a failure (learned cap, only ever lowered);
- slice subsequent bulk batches at `min(bulk_size, learned_cap)`;
- in `_add_documents_with_adaptive_batching`, when a batch is larger than the learned cap, split immediately without attempting the doomed provider call.

The existing binary splitter and single-oversized-chunk re-split stay as the safety net — behaviour on error is unchanged, only the starting size gets smarter.

Scope: single file + 2 unit tests in `tests/stores/test_opensearch_vector_store.py` (existing fakes cover the error simulation).

## Possible follow-up (out of scope)

Token-budget packing (pack batches by estimated tokens instead of document count) would avoid even the first discovery failures, but requires a per-provider token limit config knob and a real token estimator — deliberately deferred.

simcariou:
Closes #2282

## Problem

On large-document ingestion, every `bulk_size`=1000 slice was sent as-is to the embedding provider, which rejects it (400, code 3210 — Mistral limits **total tokens per request**). The adaptive splitter (from #1460) recovered by halving 1000 → 500 → 250, but remembered nothing: the next slice restarted at 1000, and the right half of a binary split re-attempted a size the left half had just seen fail. On a ~40k-chunk document that's ~100 doomed provider calls and several minutes of pure latency per ingestion.

## Fix

`OpenSearchVectorStoreAdapter` now records, on the instance, the size that triggered a split (`_learned_embedding_batch_cap`, only ever lowered):
- `add_documents` slices at `min(bulk_size, learned_cap)`, re-read each iteration;
- `_add_documents_with_adaptive_batching` pre-splits any batch above the cap without attempting the provider call.

The binary splitter, single-oversized-chunk re-split, and transient-error backoff are untouched — behaviour on error is unchanged, only the starting size gets smarter. Net effect: 2-3 discovery failures **per process lifetime** instead of per slice.

## Concurrency note

The store is a process-wide singleton reached from the FastAPI threadpool and the Temporal worker loop. The cap's check-then-set is deliberately lock-free: the worst interleaving transiently keeps the higher of two caps, costing at most one extra rejected call which re-lowers it (int assignment is GIL-atomic, no torn state). The cap is pod-local by design — each replica re-learns for the price of 2-3 calls; correctness never depends on sharing.

## Out of scope (pre-existing, worth its own issue)

`fast_store_vectors` (`features/scheduler/activities.py:218`) is an `async def` activity calling the fully synchronous `add_documents` chain (sync HTTP + `time.sleep` backoff) inline on the worker event loop — the `asyncio_0` thread visible in the logs. This change shortens time spent there but does not fix the pattern (neighbouring `output_process` already offloads via `asyncio.to_thread`).

## Tests

- Updated the existing splitting test: the doomed 3-doc re-attempt disappears (that's the improvement).
- `test_..._reuses_learned_embedding_batch_size_across_bulk_slices`: only the first slice pays a failed call.
- `test_..._does_not_reattempt_a_batch_size_that_just_failed`: size 4 attempted exactly once.

`make test`: 229 passed. `make code-quality`: clean. `fred-performance-reviewer` run: no new latency/blocking, no new metric needed, benign self-healing race documented above.

### DIFF
--- a/apps/knowledge-flow-backend/knowledge_flow_backend/core/stores/vector/opensearch_vector_store.py
+++ b/apps/knowledge-flow-backend/knowledge_flow_backend/core/stores/vector/opensearch_vector_store.py
@@ -317,6 +317,10 @@ class OpenSearchVectorStoreAdapter(BaseVectorStore):
         self._secure = secure
         self._verify_certs = verify_certs
         self._bulk_size = bulk_size
+        # Largest batch size the embedding provider has accepted after rejecting a
+        # bigger one (None until a batch-limit error is seen). Only ever lowered, so
+        # later slices skip the provider calls that are known to fail.
+        self._learned_embedding_batch_cap: Optional[int] = None
         self._embedding_model_name = embedding_model_name
         self._kpi = kpi
         self._vs: OpenSearchVectorSearch | None = None
@@ -724,6 +728,23 @@ class OpenSearchVectorStoreAdapter(BaseVectorStore):
         )
         return prepared_documents, prepared_ids
 
+    def _effective_embedding_batch_size(self) -> int:
+        if self._learned_embedding_batch_cap is None:
+            return self._bulk_size
+        return max(1, min(self._bulk_size, self._learned_embedding_batch_cap))
+
+    def _lower_learned_embedding_batch_cap(self, new_cap: int) -> None:
+        current = self._learned_embedding_batch_cap
+        if current is not None and current <= new_cap:
+            return
+        self._learned_embedding_batch_cap = new_cap
+        logger.info(
+            "[VECTOR][OPENSEARCH] learned embedding batch cap for index=%s lowered to %s (was %s)",
+            self._index,
+            new_cap,
+            current,
+        )
+
     def _add_documents_with_adaptive_batching(
         self,
         documents: List[Document],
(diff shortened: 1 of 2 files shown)


simcariou:
Added in 9f0ad705c, following live testing: **upward re-probe of the learned cap**.

A dense early batch could pin the cap lower than the rest of the corpus needs (observed live: a token-dense synthetic document settled the cap at 125 where a lighter corpus fits 250). Now, after `EMBEDDING_CAP_PROBE_AFTER_SUCCESSES` (10) consecutive successes at the cap, one slice probes `min(cap × 2, bulk_size)`:
- success → the bigger size is adopted and the streak resets;
- rejection → the cap stays, the probe batch is split as usual, the streak resets.

Worst case is one rejected call per 10 successful batches (~5-10s amortized); the cap now converges toward the real provider limit in both directions. Two tests cover the successful-probe and failed-probe paths (probe threshold monkeypatched to 2).

Full suite: 825 passed, code-quality clean.

PR #2319, opened 2026-08-10
feat(CONTENT-URL-STRATEGY): application-signed proxy URLs when presigning is unavailable

### LINKED ISSUE #2318: [CONTENT-URL-STRATEGY] Application-signed proxy URLs when presigning is unavailable (GCS without signBlob, local mode)
## Context

Browser-facing presigned URLs only work on MinIO/S3. On a private GCP deployment using GCS we do not hold `iam.serviceAccounts.signBlob` on a signing service account (request pending), and GCS V4 signing requires either that permission or an exported SA key. The platform needs to work in the meantime, and the workaround must be removable without code changes once the permission lands.

Two things are broken today:

1. **KF markdown images.** `content_controller.py:104` rewrites in-document media refs to presigned URLs so `` can load them without a bearer token. Line 111 guards on `isinstance(content_store, MinioStorageBackend)`, and KF's GCS store raises `NotImplementedError` (`gcs_content_store.py:322`). On GCS the rewrite is skipped and the unauthenticated `` request is refused.
2. **Control-plane will not start.** `app/context.py:258-266` fails fast on a GCS content store without `signing_service_account_email`. Team avatars (`teams/service.py:1517`) are the only consumer.

Same gap in local mode — `local_content_store.py:100` and `filesystem_content_store.py:420` both raise, so team avatars have never worked on a local filesystem backend.

Not a new direction: the GCS store's own docstring already records the intended fallback ("the deployment default relies on application-level HMAC download tokens (backend-agnostic)", `gcs_content_store.py:328-329`) and the HMAC primitive exists in `features/filesystem/download_token.py`. It was never built.

## Approach

RFC: `docs/swift/rfc/CONTENT-URL-STRATEGY-RFC.md`

Replace the presigned URL with an application-signed URL pointing at a minimal read-through proxy in front of the content store the app already has. The signature **is** the authorization decision, minted by code that has already run its ReBAC check — deliberately the same security model as a presigned URL, so both modes are transparently interchangeable for every consumer.

- `url_strategy: "presigned" | "proxy"` on each app's `content_storage` discriminated union, default `presigned`. No `auto` fallback — misconfiguration fails fast.
- A resolver replaces direct `get_presigned_url()` calls; the `isinstance` check at `content_controller.py:111` goes away.
- Token `{expiry}.{HMAC-SHA256(key|expiry)}`, stateless. **uid deliberately not bound** — unlike `/fs/download`, the proxy has no session to recompute it from.
- Signed-token helper + streaming router factory in `fred-core`, **mounted only when `url_strategy == "proxy"`** so no dormant unauthenticated route exists in `presigned` mode.
- TTL 60s for both media and avatars (avatars drop from 1h).
- Frontend untouched in both modes. Local mode gains working avatars as a side effect.

## Rejected (detail in RFC §3)

- Re-checking ReBAC in the proxy — needs a second key→resource→permission mapping that drifts, and breaks mode transparency.
- Frontend blob fetch — touches every consumer, breaks Range, hard to remove later.
- Scoped cookie — cookie auth in a bearer-only stack, CSRF surface.
- One central proxy in control-plane — would need credentials for every app's bucket.

## Security note

The HMAC primitive is reused but its **role is inverted**. At `mcp_fs_controller.py:230` the token is optional and layered on session auth + ReBAC — it can only narrow access. Here it is the sole grant. So: the signing secret becomes a credential (dev fallback at `download_token.py:44` should hard-fail in `proxy` mode), secrets stay per-app, and key resolution must be traversal-safe.

Pre-existing leak fixed in the same change: `content_controller.py:124` logs the signed URL at INFO and `:212` logs the whole rewritten markdown body. The tabular path already redacts (`tabular/service.py:143-145`).

## Open questions (need sign-off before implementation)

1. Resolver placement across the two store abstractions (`fred_core.store.ContentStore` vs KF's `BaseContentStore`) — recommend a standalone fred-core service, not a mixin.
2. Hard-fail on missing signing secret in `proxy` mode? — recommend yes.
3. Exclude the proxy route from the OpenAPI schema? — recommend yes (`include_in_schema=False`).
4. Raw vs opaque object key in the URL — recommend raw.

## Exit

Once `signBlob` is granted: set `url_strategy: presigned`, redeploy. No code, frontend, or contract change.

florian-muller:
Closes #2318. Design: `docs/swift/rfc/CONTENT-URL-STRATEGY-RFC.md`.

## Why

Browser-facing presigned URLs only work on MinIO/S3. On our private GCP deployment we do not hold `iam.serviceAccounts.signBlob` on a signing service account (request pending), and GCS V4 signing requires that permission or an exported SA key. Today this means **every control-plane team read fails with a 500 on GCS** (the signing-SA guard in `app/context.py` fires on first content-store use — the app boots, then breaks on `GET /teams`) and KF markdown media rewriting is silently skipped. Local filesystem mode has the same gap — team avatars have never worked there.

## What

A `url_strategy` field on each app's `content_storage` config selects between a real presigned URL and an **application-signed URL** served by a minimal read-through proxy in front of the store the app already has.

The signature **is** the authorization decision, minted by code that has already run its ReBAC check — deliberately the same security model as a presigned URL (a time-limited bearer capability over one object key), so both modes are transparently interchangeable. Frontend is untouched and byte-identical in either mode.

```yaml
content_storage:
  type: gcs
  url_strategy: proxy # "presigned" (default) | "proxy"
```

- **fred-core**: `signed_token.py` (HMAC-SHA256, components bound as a sequence), `content_url_resolver.py` (standalone service, not a store mixin), `object_proxy_router.py` (`include_in_schema=False`, Range/206, `Cache-Control: private` + `ETag`, traversal-safe keys).
- **Both apps**: config field, conditional mount (no dormant unauthenticated route in `presigned` mode), resolver wired into KF markdown media and CP team avatars.
- The GCS signing-SA guard now applies **only** in `presigned` mode — that is the fix that makes control-plane usable on GCS.
- `isinstance(content_store, MinioStorageBackend)` in `content_controller.py` is gone; backend capability was never a type question.
- Avatar TTL 1h to 60s, matching media.
- `get_presigned_url_internal` (tabular/DuckDB) deliberately untouched.

## Security

**The HMAC primitive's role is inverted here.** At `mcp_fs_controller.py:230` the `/fs/download` token is optional and layered on session auth + ReBAC — it can only *narrow* access. In the object proxy the token is the *sole* grant. Consequently:

- **The hardcoded dev signing key is removed outright** (was `download_token.py:_DEV_FALLBACK`). A default signing key in shipped source is unsafe regardless of which route consumes it. Missing secret now raises with a message naming both the variable and `config/.env`.
- Signing secrets are **per-app** — a KF token must not verify against control-plane objects.
- Object keys are traversal-checked; there is no ReBAC check downstream.
- **Fixes a pre-existing credential leak**: `content_controller.py` logged the generated signed URL at INFO and the entire rewritten markdown body. Both removed, following the redaction precedent in `tabular/service.py`.

## Reviewer attention

1. **KF markdown images will still not render, for a separate pre-existing reason.** The rewrite has always built key `{uid}/output/media/{file}` and resolved it against the **`-objects`** bucket, while `save_output`/`get_media` write and read media in the **`-documents`** bucket (`minio_content_store.py:227` vs `:535`; same split in all three KF stores). So it has been pointing at a non-existent object on MinIO too. This PR fixes the *mechanism* and faithfully inherits the wrong key — it does not fix the symptom. Not patched silently because the fix is a design choice (document-area read path on `BaseContentStore`, or change the key). Follow-up issue recommended; see RFC section 8.3.
2. **Tabular SQL preview stays unavailable on GCS.** `get_presigned_url_internal` always signs internally, so `proxy` lets the pod boot but DuckDB Parquet reads still fail at query time. Called out in the config description, `values-gcp.yaml` and RFC section 8.4.
3. **`KNOWLEDGE_FLOW_DOWNLOAD_SECRET` is now mandatory.** Removing the dev fallback means any KF deployment or local run without it hard-errors on the first `/fs/download` mint/verify. Added to `.env.template`, `values-gcp.yaml` and `ENV_VARIABLES.md` — **existing environments must set it before rollout**.
4. **Bandwidth**: avatar/media bytes traverse the API pod in proxy mode. Fine for these two consumers; large PDFs must stay on the bearer-authenticated streaming path.

## End-to-end verification

Beyond unit tests, the feature was exercised against a full local stack (postgres, keycloak, openfga, opensearch, seaweedfs) with real uvicorn servers driven by curl — no TestClient.

- **The core case**: avatar uploaded to local filesystem storage in `proxy` mode, then fetched with a **plain unauthenticated GET** — 200, 12420 bytes, sha256 identical to the upload. This is the case that has never worked on local storage.
- **Token enforcement**: nine attack shapes all 403 — absent, tampered, minted for another key, minted with the wrong secret, expiry bumped while keeping the signature, expired, garbage. Natural 60s expiry confirmed live (200 at t+2s, 403 at t+65s). Body never distinguishes expired from forged.
- **Traversal**: ten variants incl. URL- and double-encoded, all 400/403; no response body ever contained `root:x:`.
- **Range**: suffix ranges, clamping and 416 all correct. **Headers**: `cache-control: private, max-age=<remaining token life>`, `ETag`.
- **`presigned` mode**: route 404s, absent from `openapi.json` in every mode, MinIO still mints real presigned URLs.
- **Cross-app isolation**: a control-plane token replayed against knowledge-flow → 403. **Log hygiene**: tokens redacted by the existing uvicorn filter.
- **KF media**: URL minted, token verifies, then 404 — proven to be the bucket mismatch by copying the object into the `-objects` bucket and re-fetching (200, correct bytes, Range working). Mechanism sound; remaining bug is item 1 above.

Three issues found by that pass are fixed in this PR: `HEAD` on a proxy URL returned 405 where a real presigned URL answers it (mode-parity divergence); the `context.py` comment claiming a startup failure was inaccurate (corrected above); and `teams/service.py` rebuilt the URL resolver per request instead of using the cached accessor, re-reading config and the signing secret from env on every team read.

Two findings deliberately not fixed: `Cache-Control` buys less than hoped because the token is re-minted per read so the browser cache key changes each time (matches spec, design note), and `_MEDIA_RE` mis-captures inline-style markdown image links — pre-existing, harmless today because the only producer emits reference-style links.

## Verification

| Project | `make code-quality` | `make test` |
|---|---|---|
| libs/fred-core | 0 errors | 488 passed |
| apps/knowledge-flow-backend | 0 errors | 846 passed, 2 failed |
| apps/control-plane-backend | 0 errors | 721 passed |

The 2 KF failures are pre-existing and unrelated (`test_prometheus_service.py`, cross-test env pollution on `PROMETHEUS_BEARER_TOKEN`/`PROMETHEUS_PASSWORD`); they reproduce on a clean tree and pass in isolation. `make check-config-files`, `make check-chart-values`, `make check-chart-schema-drift` all pass.

**OpenAPI**: regeneration verified as a genuine no-op, not assumed — both specs were regenerated and inspected. No `/objects` path (route is schema-excluded and unmounted in `presigned` mode) and `url_strategy` appears nowhere (config models are not API models). No frontend change.

## Exit

Once `signBlob` is granted: set `url_strategy: presigned`, redeploy. No code, frontend, or contract change.


simcariou:
Ah oui tu es parti deep

florian-muller:
> Ah oui tu es parti deep
> 

Haha ouai, j'avais une idée, je voulais tenter, mais j'ai pas eu le temps de tester avant de partir.

Si ça te semble une bonne idée et que ça marche, tu peux merger. 

L'idée c'est d'avoir un paramètre qui fait qu'on gère les url pré signé nous même, pour by pass temporairement le problème de droit qu'on a un prod. C'est moins bien par ce que du coup on stream les fichiers par nos serveurs, mais au moins ça devrait marcher. L'intérêt c'est que ça reprend la même interface que des url pré signé classique donc le reste du code change pas.

dimitri-tombroff:
i close this I believe this is obsolete. Please reopen if i am wrong

Issue #2378, opened 2026-08-14
[Design] Define safe durable managed-chat history refresh after discarded overlap


adriang-tech:
## Context

Managed chat discards a session-history response when the request overlaps live chat activity. The response is neither applied nor cached, because it can omit an in-flight or recently completed exchange.

The client intentionally does not retry that discarded response automatically. Persistence may still lag after the live turn ends, so blindly replacing the thread with a later full-history response can still overwrite newer local state.

This issue defines the durable convergence mechanism separately from #2372.

When Fred opens an existing conversation, the browser asks the server for its saved messages. That request takes some time. If the user sends a new message before the request finishes, the response may describe the conversation as it existed before that new message.

Replacing the visible conversation with that older response could make the user's new message or an in-progress answer disappear. Fred therefore discards a history response that overlapped live chat activity. This is safe, but Fred still needs a reliable way to synchronize with saved history afterward.

### Simple example

1. A user opens an existing conversation.
2. Fred starts loading its saved history.
3. Before loading finishes, the user sends “What is our holiday policy?”
4. The older history response arrives without that new message.
5. Replacing the screen with this response would make the new question disappear.
6. Fred discards the older response instead.
7. Fred must later refresh safely, after it can prove the saved history includes the completed turn.

The race can happen even when loading takes only a few hundred milliseconds; it does not require an unusually slow request.

## Design goal

Define and implement a safe way for the active session to converge to durable history after an overlapping response is discarded, without replacing newer local messages or replaying stale cache data.

The design should evaluate approaches such as:

- temporarily blocking all chat and HITL submissions while authoritative history is loading;
- merging/upserting persisted messages into the current thread with explicit message identity and ordering rules;
- server-provided history revisions, cursors, or freshness metadata;
- another mechanism that proves a response is authoritative before full replacement.

### Lowest-complexity option to evaluate

Fred could allow typing while history loads but disable submission until the authoritative history request finishes. This must be enforced in the send handlers, not only by disabling the visible button, so Enter-to-send, fixed HITL choices, free-text HITL answers, and programmatic calls cannot bypass it.

This option also needs a clear loading state and retry/recovery behavior when history loading fails; otherwise a conversation could remain unusable. It can eliminate the initial load-versus-send race, but it does not by itself resolve every synchronization case, such as background refreshes or changes made in another browser tab. The design should decide whether those cases are required before introducing a more complex merge or revision protocol.

## Required properties

[...]

Issue #2372, opened 2026-08-14
[Bug+Design] Follow-up to #2253: harden managed-chat rejection recovery, concurrent state, and durable history convergence


adriang-tech:
## Context

Issue #2253 introduced the deployment-configurable chat-input character limit and frontend enforcement.

Several frontend race conditions and recovery gaps remain around runtime rejection handling, session-history refreshes, attachments, and HITL prompts.

This follow-up hardens those behaviors without changing the character-counting policy or backend limit contract.

### Runtime error handling

- Read at most 16 KiB from unsuccessful runtime responses.
- Cancel oversized or unread response bodies.
- Handle locked response streams without losing draft recovery.
- Bound displayed error messages by Unicode code points.
- Support these structured error shapes:
  - `detail`
  - `detail.message`
  - FastAPI validation arrays
  - `errors[].detail` and `errors[].message`
  - top-level `message`
- Keep non-JSON, oversized, or unrecognized responses generic.

### Runtime-reported limit handling

- Treat `chat_input_too_long.limit_chars` as authoritative.
- Prevent an older in-flight preparation response from overwriting a newer rejection.
- Allow a later valid preparation to raise or replace the displayed limit.
- Preserve a rejection-learned limit when a legacy runtime omits `max_chat_input_chars`.
- Ignore invalid, non-positive, or non-integral limit metadata.
- Preserve the last known limit across session-local resets.

### Draft and attachment recovery

- Restore the complete ordinary-chat draft after a length rejection.
- Restore exactly the ready attachments removed when the turn started.
- Preserve inline-image context.
- Do not restore attachments deleted while the request was in flight.
- Include uploads that became ready during asynchronous send preparation.
- Release retained rollback state after the runtime accepts the request.
- Restore displaced HITL state when ordinary-send preflight fails or is explicitly stopped.

### HITL recovery

- Preserve controlled HITL free-text drafts.
- Do not count or submit an unrelated free-text draft when a fixed choice is selected.
- Restore the prompt when a resume never reached the runtime.
- Do not restore a consumed prompt after a successful runtime response.
- Do not let a superseded resume restore its old prompt over a newer ordinary turn.
- Keep newer HITL prompts authoritative.
- Avoid state-ref mutation inside React state updaters.

### Session-history concurrency

- Discard, without applying or caching, any history response whose request overlapped live chat activity.
[...]

Issue #2253, opened 2026-08-05
Limit chat input message length


simcariou:
## Goal

Add a configurable maximum length for one submitted chat message. The runtime backend is authoritative, while the managed-chat frontend shows the same limit and prevents an oversized draft from being sent.

The default limit is **5,000 Unicode code points**. A message containing exactly 5,000 code points is accepted; 5,001 is rejected.

## Requirements

- Define `app.max_chat_input_chars` in the agent-pod configuration, with a default of `5000` and startup validation requiring a positive integer.
- Expose the value in the checked-in development and production YAML files and as `applications.fred-agents.configuration.app.max_chat_input_chars` in the Helm values.
- Enforce the effective value in the runtime backend. Frontend enforcement is user experience, not the security boundary.
- Cover the current submitted message on:
  - `POST /agents/execute`
  - `POST /agents/execute/stream`
  - `POST /agents/evaluate`
  - `POST /v1/chat/completions`, applied to the last user message Fred forwards
  - HITL resumes represented as a bare string or string-valued `choice_id`, `answer`, and `text` fields
- For a dict-shaped HITL resume, count the combined code points in the canonical `choice_id`, `answer`, and `text` values. Do not recursively interpret arbitrary agent-defined JSON keys as chat text.
- Use matching counting semantics in both languages:
  - Python: `len(text)`
  - TypeScript: allocation-free `for...of` code-point iteration equivalent to `[...text].length`, not JavaScript `string.length`
- Do not silently truncate the message, use native textarea `maxLength`, or discard the user's draft.
- Do not include submitted message text in errors, logs, metrics, or traces.
- Reject at the beginning of each runtime handler, before exchange lookup, target resolution, history persistence, stream construction, or agent execution.

## Error contracts

Fred-native execution endpoints return HTTP `422`:

```json
{
  "detail": {
    "code": "chat_input_too_long",
    "message": "Your message exceeds the 5,000-character limit.",
    "limit_chars": 5000,
    "actual_chars": 6127
  }
}
```

The OpenAI-compatible endpoint returns HTTP `400` using its `invalid_request_error` envelope, with `param: "messages"` and `code: "chat_input_too_long"`.

The response must report only the configured limit and actual code-point count; it must never echo the rejected content.

### Runtime configuration and publication

- Add `max_chat_input_chars: int = Field(default=5_000, ge=1, ...)` to `PodAppConfig` in `libs/fred-runtime/fred_runtime/app/config.py`.
[...]

Issue #2221, opened 2026-08-04
Chat composer input lag on long conversations: full message-thread re-render + markdown re-parse per keystroke


dimitri-tombroff:
## Problem

Observed live in production on a shared screen: in a long-running conversation, typing a single character in the chat composer and hitting Enter took several seconds to register. Severity scales with conversation length.

## Root cause (confirmed by code read, not yet reproduced live)

1. The composer's `input` state lives in `useManagedChat()`, called directly at the top of `ManagedChatPage` (`apps/frontend/src/rework/components/pages/ManagedChatPage/ManagedChatPage.tsx`). Every `setInput()` call — i.e. every keystroke — re-renders the whole `ManagedChatPage` component tree.
2. `ConversationThread`, `AssistantTurn`, `UserTurn`, and `MarkdownRenderer` have no `React.memo` anywhere (`apps/frontend/src/rework/components/pages/ManagedChatPage/ConversationThread/ConversationThread.tsx`, `apps/frontend/src/rework/components/shared/organisms/AssistantTurn/AssistantTurn.tsx`, `apps/frontend/src/rework/components/shared/molecules/MarkdownRenderer/MarkdownRenderer.tsx`). Even though `threadMessages` itself is a stable memoized reference (`useMemo` in `useManagedChat.ts`), React still walks down and re-renders every child on each keystroke since there is no memo boundary anywhere in that subtree.
3. Each historical assistant message re-runs the full `MarkdownRenderer` pipeline — `react-markdown` + `remark-gfm` + `remark-math` + `remark-directive` + `rehype-katex` + `rehype-sanitize` — none of it memoized on `text`. For N past messages, that is N full markdown-parse-and-sanitize passes fired synchronously on the main thread on every character typed.

This explains why the lag is invisible on short conversations and grows with thread length: the composer keystroke handler's cost is O(number of historical messages), not O(1).

Recent chat-page polish (#2218, chain-of-thought redesign + per-step token usage, merged 2026-08-04) did not introduce this — the missing memo boundary predates it — but it adds more per-message render work sitting behind the same unmemoized path, compounding the effect going forward.

## Suggested fix direction (not yet applied)

- Wrap `ConversationThread` (or at minimum `AssistantTurn`/`UserTurn`) in `React.memo`.
- Memoize `MarkdownRenderer` on `text`/`streaming` so historical messages don't re-parse when unrelated state changes.
[...]

PR #2381, opened 2026-08-17
fix(#2372): harden managed chat rejection recovery

adriang-tech:

**Issue:** #2372 (follow-up to #2253). Durable history convergence split out to #2378.

**Problem in one sentence:** Managed chat could lose or replace an active draft, its attachments, a displaced HITL prompt, or thread history when a runtime rejection overlapped asynchronous send preparation, and runtime error bodies were consumed without bounds.

---

**RFC written?** Not required — this is hardening of behaviour already specified by #2253, with the required changes enumerated in #2372 itself. The one genuine design decision (dropping automatic history retry) is recorded in #2372's amended body and split into #2378 rather than a separate RFC.

---

## 3. What you built

- **Runtime errors** — read at most 16 KiB from an unsuccessful response, cancel oversized or unread bodies, survive a locked stream without losing draft recovery, bound displayed messages by Unicode code points, and parse only supported envelopes (`detail`, `detail.message`, FastAPI validation arrays, `errors[].detail`/`errors[].message`, top-level `message`). Anything else stays generic, so rejected content cannot reach logs or toasts.
- **Limit handling** — `chat_input_too_long.limit_chars` is authoritative; an older in-flight preparation cannot overwrite a newer rejection; a later valid preparation may still raise the limit; invalid, non-positive and non-integral metadata is ignored.
- **Draft and attachment recovery** — each send owns its rollback explicitly. Attachment context resolves at the final turn-start boundary, so a file that becomes ready during preparation joins that turn and a later one is retained for the next. A rejection restores the exact rejected draft without overwriting newer edits.
- **HITL and history** — a superseded resume cannot replace a newer prompt; fixed-choice and free-text behaviour are preserved; a history snapshot cannot replace live activity. The #2221 conversation-rendering boundary is retained.

### Scope decision reviewers should weigh

#2372 originally required **retrying** history responses discarded because of live activity. That was deliberately reversed: persistence can lag a completed turn, so a later full-thread replacement is not inherently more authoritative and retrying risks overwriting newer local turns. This branch discards without retry, the issue body and both contracts were amended to match, and durable convergence is tracked in #2378. Agree with the reversal or say so — it is the one substantive design change here.

---

## 4. Proof of quality

**Tests added or updated:** this branch adds race-discriminating coverage across acceptance, rejection, limits, attachments, HITL, history, loading and rendering — 20 files, +2,931/−281. Rather than list every case, the coverage was verified by a counterfactual (mutation) pass: **42 mutations** applied across error handling, limit adoption, draft/attachment rollback, HITL guards and history discard. **Every one produced a failing test**; none passed silently.

Mutations that each proved a distinct guard include: the 16 KiB byte cap; `discard()`/`reader.cancel()` on unread bodies; code-point (not UTF-16) truncation; all six error-shape branches; both halves of `positiveInteger`; the `err.status === 422` recovery gate; whitespace-exact draft restore; attempt-owned delete rollback; the superseded-resume and newer-prompt HITL guards; and all three history-discard limbs.

```
npx tsc --noEmit
npx prettier . --check
Checking formatting...
All matched files use Prettier code style!
npx eslint "src/**/*.{ts,tsx}"
All frontend code quality checks completed
```

Root `make code-quality` was **not** run: it loops the eight Python packages plus frontend, and this branch changes **0 `.py` files** (11 tsx, 7 ts, 2 md), so the Python legs are identical to base. `basedpyright` n/a for the same reason.

**`make test` output:**

```
Test Files  137 passed | 1 skipped (138)
Tests       1427 passed | 2 skipped (1429)
```

Green with no exceptions. An earlier revision of this branch carried 2 inherited `DocumentUploadDrawer.maxDepth` failures; rebasing onto current `swift` picked up their fix, so that caveat no longer applies.

One gap worth stating rather than hiding:

- The two changed `docs/**/*.md` files fail `prettier`. They fail **identically on `origin/swift`**, and `docs/` is outside `apps/frontend`, so no gate covers them. Not a regression (re-verified after the rebase).
- **No CI workflow runs the frontend test suite** (`vitest` appears in no workflow). Frontend `type-check` and `format-check` do run and pass. So the coverage this branch adds is not enforced by CI.

---

## 5. Docs updated

| ------------ | ------------ |
| Backlog `[ ]` item is now done | n/a — backlog frozen; tracked on #2372 |
| New behaviour, API field, or contract change | `docs/swift/design/RUNTIME-EXECUTION-CONTRACT.md` |
| Frozen contract touched (`execution.py`, `agent_app.py`, OpenAPI) | n/a — frontend-only, no backend or generated client touched |
| UX component implemented or visual status changed | `docs/swift/ux/COMPONENT-UX.md` |
| WORKPLAN sprint item finished | n/a — WORKPLAN frozen |
| Code and a design doc now diverge | No — both contracts amended in this branch to match the no-retry decision |

---

## 6. Close-out statement

```
## Task close-out
- Code: bounded runtime-error consumption, authoritative limit adoption, per-send rollback
  ownership for drafts/attachments, HITL prompt authority, and history discard-without-retry
- Tests: 1427 passed / 2 skipped / 0 failures; 42-mutation counterfactual pass, all caught
- Docs updated: RUNTIME-EXECUTION-CONTRACT.md, COMPONENT-UX.md
- Backlog: n/a (frozen) — tracked on #2372; durable history convergence split to #2378
- Skipped steps: root make code-quality (0 .py changed); Python gates and migration checks n/a
```

---

## 7. Risk and rollback

**What breaks if this is wrong?** The blast radius is the managed-chat composer. A regression in rollback ownership loses a user's draft or attachment chips on a runtime rejection; a regression in the HITL guards can render a stale prompt over a live turn, and answering it would resume against a consumed checkpoint. The history change is the highest-judgement item: discarding without retry means an overlapping response is not re-fetched while the user stays on that session — a session switch or remount does refetch, and live/local messages remain visible.

**How do we roll back?** Frontend-only, single commit, no migration and no backend or generated-client change — revert the commit and redeploy the frontend. No data or schema is affected.

### DIFF
--- a/apps/frontend/src/rework/components/pages/ManagedChatPage/ConversationThread/ConversationThread.test.tsx
+++ b/apps/frontend/src/rework/components/pages/ManagedChatPage/ConversationThread/ConversationThread.test.tsx
@@ -1,3 +1,4 @@
+// @vitest-environment happy-dom
 // Copyright Thales 2026
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -12,16 +13,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { createRef, type ReactNode } from "react";
+import { act, createRef, type ReactNode } from "react";
+import { createRoot } from "react-dom/client";
 import { renderToStaticMarkup } from "react-dom/server";
 import { describe, expect, it, vi } from "vitest";
+
+const assistantTurnRenderMock = vi.hoisted(() => vi.fn());
 
 vi.mock("react-i18next", () => ({ useTranslation: () => ({ t: (key: string) => key }) }));
 vi.mock("@shared/organisms/ChatMessagesArea/ChatMessagesArea", () => ({
   ChatMessagesArea: ({ children }: { children?: ReactNode }) => <div>{children}</div>,
 }));
 vi.mock("@shared/organisms/UserTurn/UserTurn", () => ({ UserTurn: () => null }));
-vi.mock("@shared/organisms/AssistantTurn/AssistantTurn", () => ({ AssistantTurn: () => null }));
+vi.mock("@shared/organisms/AssistantTurn/AssistantTurn", async () => {
+  const { memo } = await vi.importActual<typeof import("react")>("react");
+  return {
+    AssistantTurn: memo(function AssistantTurn(props: { pendingToolCallIds?: readonly string[] | null }) {
+      assistantTurnRenderMock(props);
+      return null;
+    }),
+  };
+});
 vi.mock("@shared/molecules/HitlPrompt/HitlPrompt.tsx", () => ({
   HitlPrompt: (props: {
     maxChatInputChars?: number;
@@ -65,5 +77,51 @@ describe("ConversationThread HITL input policy wiring", () => {
     expect(html).toContain('data-free-text="full draft"');
     expect(html).toContain('data-has-answer-handler="true"');
     expect(html).toContain('data-has-change-handler="true"');
+  });
+
+  it("does not re-render historical assistant turns when only the HITL draft changes", () => {
+    globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+    assistantTurnRenderMock.mockClear();
+    const container = document.createElement("div");
+    document.body.appendChild(container);
+    const root = createRoot(container);
+    const pendingHitl = {
+      type: "awaiting_human" as const,
+      session_id: "session-1",
+      exchange_id: "exchange-1",
+      payload: { free_text: true, pending_calls: [] },
+    };
+    const messages = [
+      {
+        id: "assistant-1",
+        role: "assistant" as const,
+        text: "historical markdown",
+        traceMessages: [],
+        sources: [],
+        uiParts: [],
+        isStreaming: false,
+      },
+    ];
+    const render = (hitlFreeText: string) => (
+      <ConversationThread
+        messages={messages}
+        pendingHitl={pendingHitl}
+        isLoading={false}
+        isStreaming={false}
+        scrollContainerRef={createRef<HTMLDivElement>()}
+        onHitlAnswer={() => undefined}
+        maxChatInputChars={5_000}
+        hitlFreeText={hitlFreeText}
+        onHitlFreeTextChange={() => undefined}
+      />
+    );
+
+    act(() => root.render(render("a")));
+    expect(assistantTurnRenderMock).toHaveBeenCalledTimes(1);
+    act(() => root.render(render("ab")));
+    expect(assistantTurnRenderMock).toHaveBeenCalledTimes(1);
+
+    act(() => root.unmount());
+    container.remove();
   });
 });
--- a/apps/frontend/src/rework/components/pages/ManagedChatPage/ConversationThread/ConversationThread.tsx
+++ b/apps/frontend/src/rework/components/pages/ManagedChatPage/ConversationThread/ConversationThread.tsx
@@ -15,7 +15,7 @@
 // Page-local composition of organisms for ManagedChatPage.
 // Lives under pages/ so it may import from shared/organisms freely.
 
-import { memo, type ReactNode, type RefObject } from "react";
+import { memo, useMemo, type ReactNode, type RefObject } from "react";
 import { useTranslation } from "react-i18next";
 import type { RuntimeAwaitingHumanEvent } from "@hooks/useChatSse";
 import { useAssistantCopyInterception } from "@hooks/useAssistantCopyInterception";
@@ -63,8 +63,10 @@ export const ConversationThread = memo(function ConversationThread({
   // #2177: one prompt can batch several calls at once) — lets the trace row
   // for each of those tools render "awaiting confirmation" instead of
   // "running" while the prompt below is still unanswered.
-  const pendingToolCallIds =
-    pendingHitl?.payload.pending_calls?.map((c) => c.tool_call_id).filter((id): id is string => !!id) ?? null;
+  const pendingToolCallIds = useMemo(
+    () => pendingHitl?.payload.pending_calls?.map((c) => c.tool_call_id).filter((id): id is string => !!id) ?? null,
+    [pendingHitl],
+  );
 
   return (
     <ChatMessagesArea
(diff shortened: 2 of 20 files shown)


adriang-tech:
<head></head><blockquote dir="ltr" style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><p dir="ltr"><strong>Closing without merging.</strong></p><p dir="ltr">The behaviours fixed here are real, but review established that each needs something else to go wrong first — none is reachable in normal use:</p><div dir="ltr">
Behaviour | What it requires
-- | --
Attachments lost on a length rejection | a legacy runtime that doesn't publish max_chat_input_chars, a mismatched replica, or paste-and-send inside one prepare-execution round trip
Unparseable 422 loses the draft | a 422 with a non-standard body — Fred's own rejection body is a fixed ~148 bytes
HITL prompt stranded | an active tool-approval gate, an ordinary send instead of an answer, and that send failing or being stopped
Unbounded error body | pre-existing, not introduced by #2253 — which actually narrowed it, from dumping the whole body to surfacing only detail.message. Unreachable from Fred's runtime
Stale history replacing a live turn | the history round trip must outlast open → send → turn completion

</div><p dir="ltr">Two things worth recording so they aren't re-derived:</p><p dir="ltr"><strong>The frontend gate blocks the ordinary case.</strong><span class="Apple-converted-space"> </span>Above the limit the send button is genuinely disabled (<code>inputTooLong</code>), so a long paste in a settled chat never reaches the runtime. The 422 path exists because #2253 documents the displayed limit as<span class="Apple-converted-space"> </span><em>advisory</em><span class="Apple-converted-space"> </span>with the runtime authoritative — not because the gate is missing.</p><p dir="ltr"><strong>The one exception is a session switch.</strong><span class="Apple-converted-space"> </span><code>reset()</code><span class="Apple-converted-space"> </span>calls<span class="Apple-converted-space"> </span><code>setMaxChatInputChars(undefined)</code><span class="Apple-converted-space"> </span>and fires on every<span class="Apple-converted-space"> </span><code>sessionId</code>change, and the gate is<span class="Apple-converted-space"> </span><code>maxChatInputChars !== undefined &amp;&amp; …</code><span class="Apple-converted-space"> </span>— so between switching conversations and the next<span class="Apple-converted-space"> </span><code>prepare-execution</code><span class="Apple-converted-space"> </span>resolving, there is no counter and no length gating. Still a single-round-trip window, but the most reachable of these.</p><p dir="ltr"><strong>Worth lifting separately:</strong><span class="Apple-converted-space"> </span>dropping<span class="Apple-converted-space"> </span><code>setMaxChatInputChars(undefined)</code><span class="Apple-converted-space"> </span>from<span class="Apple-converted-space"> </span><code>reset()</code><span class="Apple-converted-space"> </span>in<span class="Apple-converted-space"> </span><code>useChatSse.ts</code>. One deleted line, no coordination state, and it closes that window on its own. The limit is deployment-scoped and the mount is keyed on<span class="Apple-converted-space"> </span><code>agentInstanceId</code>, so clearing it per-session was always slightly wrong.</p><p dir="ltr"><strong>Preserved:</strong><span class="Apple-converted-space"> </span>the full work is at tag<span class="Apple-converted-space"> </span><code>pr2381-prerebase</code><span class="Apple-converted-space"> </span>(<code>1038c48d</code>), including the 42-mutation counterfactual coverage and the RUNTIME-EXECUTION-CONTRACT / COMPONENT-UX updates.<span class="Apple-converted-space"> </span><code>wip/2253-preclean</code>holds an alternative approach where history<span class="Apple-converted-space"> </span><em>merges</em><span class="Apple-converted-space"> </span>into the thread via<span class="Apple-converted-space"> </span><code>upsertOne</code><span class="Apple-converted-space"> </span>instead of replacing it — relevant to #2378, since that removes the need for the staleness guards entirely.</p><p dir="ltr"><a href="https://github.com/ThalesGroup/fred/issues/2372" target="_blank" rel="noreferrer">#2372</a><span class="Apple-converted-space"> </span>and<span class="Apple-converted-space"> </span><a href="https://github.com/ThalesGroup/fred/issues/2378" target="_blank" rel="noreferrer">#2378</a><span class="Apple-converted-space"> </span>remain open.</p></blockquote><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Closing without merging.

The behaviours fixed here are real, but review established that each needs something else to go wrong first — none is reachable in normal use:

Behaviour	What it requires
Attachments lost on a length rejection	a legacy runtime that doesn't publish max_chat_input_chars, a mismatched replica, or paste-and-send inside one prepare-execution round trip
Unparseable 422 loses the draft	a 422 with a non-standard body — Fred's own rejection body is a fixed ~148 bytes
HITL prompt stranded	an active tool-approval gate, an ordinary send instead of an answer, and that send failing or being stopped
Unbounded error body	pre-existing, not introduced by [#2253](https://github.com/ThalesGroup/fred/issues/2253) — which actually narrowed it, from dumping the whole body to surfacing only detail.message. Unreachable from Fred's runtime
Stale history replacing a live turn	the history round trip must outlast open → send → turn completion

Two things worth recording so they aren't re-derived:

The frontend gate blocks the ordinary case. Above the limit the send button is genuinely disabled (inputTooLong), so a long paste in a settled chat never reaches the runtime. The 422 path exists because #2253 documents the displayed limit as advisory with the runtime authoritative — not because the gate is missing.

The one exception is a session switch. reset() calls setMaxChatInputChars(undefined) and fires on every sessionId change, and the gate is maxChatInputChars !== undefined && … — so between switching conversations and the next prepare-execution resolving, there is no counter and no length gating. Still a single-round-trip window, but the most reachable of these.

Worth lifting separately: dropping setMaxChatInputChars(undefined) from reset() in useChatSse.ts. One deleted line, no coordination state, and it closes that window on its own. The limit is deployment-scoped and the mount is keyed on agentInstanceId, so clearing it per-session was always slightly wrong.

Preserved: the full work is at tag pr2381-prerebase (1038c48d), including the 42-mutation counterfactual coverage and the RUNTIME-EXECUTION-CONTRACT / COMPONENT-UX updates. wip/2253-preclean holds an alternative approach where history merges into the thread via upsertOne instead of replacing it — relevant to #2378, since that removes the need for the staleness guards entirely.

[#2372](https://github.com/ThalesGroup/fred/issues/2372) and [#2378](https://github.com/ThalesGroup/fred/issues/2378) remain open.

PR #2437, opened 2026-08-26
impr(#2443): New teams are now private by default

### LINKED ISSUE #2433: feat(TEAM-10): new teams default to private visibility
## Problem

Every team created today lands with `visibility: public` (contract note
CONTROL-PLANE-PRODUCT-CONTRACT.md §30) - it is immediately listed in the
marketplace "Discover" section for every user, and readable by non-members
through the ReBAC `public` relation, unless a team admin remembers to flip
it to private afterwards. On prism-c2-prod this means the marketplace
lists many teams whose admins never intended broad discoverability.

The product stance changes: **a new team must start private** - invisible
to non-members until its admin deliberately opts into the marketplace.

## Change (new teams only)

- `TeamMetadataRow.visibility` ORM default `"public"` -> `"private"`
  (fred-core; `TeamMetadataStore.create` relies on the column default).
- Align the Pydantic defaults: `TeamMetadata.visibility` (fred-core) and
  `Team.visibility` (control-plane response schema) -> regenerate
  `openapi.json` (+ frontend client if it diffs).
- `create_team` currently grants the ReBAC `public` relation
  unconditionally right after bootstrap (service.py:572) - make it
  conditional on the created team's visibility, otherwise a brand-new
  private team stays readable/discoverable until the next `_list_teams`
  lazy backfill.
- New linear Alembic migration: `ALTER COLUMN visibility SET DEFAULT
  'private'` (server_default alignment, no data touched).
- Frontend fallback `team.visibility ?? "public"` -> `?? "private"`
  (TeamSettingsParameters).
- Importer keeps defaulting pre-visibility bundles to `public` (fidelity
  to the exported platform's behavior) - documented, not changed.

## Explicitly out of scope

**Existing teams keep their stored `public` value** - nothing becomes
private as a side effect of the rollout (decided 2026-08-26). Hiding a
pre-existing team remains a per-team admin action in its settings.

## Docs

- Dated contract note in `CONTROL-PLANE-PRODUCT-CONTRACT.md` (supersedes
  the "default `public`" statement of §30).
- `RUNTIME-EXECUTION-CONTRACT.md` line ~1727 says PUBLIC is "the default
  for every team" - retouch.

## Tests

- fred-core: new team metadata defaults to private.
- control-plane: `create_team` does not grant the `public` relation for a
  default (private) team; still grants it for an explicit public one.
- Adjust any test asserting the old public default.

simcariou:
Closes #2433.

## What

A brand-new team now starts **private** - invisible to non-members (no marketplace listing, `GET /teams/{id}` 403s for them) until a team admin deliberately flips it to public in the team settings.

## How

- `TeamMetadataRow.visibility` ORM default `"public"` -> `"private"` (fred-core). This is the governing default: `TeamMetadataStore.create` inserts only `(id, name)`. The `TeamMetadata` and `Team` pydantic defaults are mirrored so the OpenAPI spec tells the truth.
- `create_team`'s immediate TEAM-09 `public`-relation grant is now **conditional** on the created metadata's visibility - a default (private) team never holds the ReBAC `public` tuple, not even transiently (a grant-then-lazy-revoke would leave it readable by anyone until the next `_list_teams` pass). The idempotent grant/revoke backfill in `_list_teams` is unchanged and remains the backstop.
- Alembic `0c70cb820802` aligns the DB `server_default` (batch mode for SQLite, same pattern as `9ee7b44b0d57`). Single linear head, verified with a real upgrade -> downgrade -> upgrade cycle.
- `build_personal_team` states `visibility: "private"` explicitly (it previously inherited the schema default and reported `"public"`) - truthful for a space that is never marketplace-listed and never readable by non-members.
- Frontend: the `TeamSettingsParameters` absent-field fallback mirrors the new default. Client regenerated via `cd apps/frontend && make update-control-plane-api` - no TS diff (property defaults don't materialize in the generated client).

## Explicitly out of scope

**Existing teams keep their stored `public` value** - nothing becomes private as a side effect of this rollout (decided 2026-08-26, see #2433). Hiding a pre-existing team remains a per-team admin action in its settings. Same fidelity rule on import: pre-visibility bundles still import as `public`.

## Tests

- fred-core: new SQLite-backed test proving a real `create()` row lands `private`/`invite_only` (exercises the ORM default, not a pydantic-level one).
- control-plane: `create_team` default -> `PRIVATE` response + zero `public` writes; explicit-public metadata -> grant still happens; personal-team contract tests updated to the explicit `"private"`; the `update_team` permission-gating test now seeds an explicitly public team so the TEAM-10 downgrade rule (which has its own tests) stays out of its way.
- Suites: control-plane 851 passed, fred-core offline 521 passed, frontend 1418 passed, root `make code-quality` green.

## Docs

- `CONTROL-PLANE-PRODUCT-CONTRACT.md` new §44 (supersedes §30's "Default and migration" paragraph).
- Dated retouches where "public is the default for every team" was stated: `RUNTIME-EXECUTION-CONTRACT.md` (x2), `COMPONENT-UX.md`.

## Note for reviewers

`FRED-TEAM-CONFIG-RFC.md` §5.1.2 still says `PUBLIC # default` - that RFC records the TEAM-10 proposal as written (2026-07-26) and its settled content already lives in the contract doc; per the RFC-vs-doc rule the trim/archive of that section is left to a separate consolidation change rather than bundled here.


simcariou:
## Review pass (8 angles) - applied in d239baab4

No correctness bug was found in the original diff. Findings that mattered, and what was done:

**Fixed here**
- `backfill_storage_usage.py` (knowledge-flow) manufactures a `TeamMetadataRow` for a team that pre-dates its registry row, relying on the ORM default - after the flip such a team would land private and `_list_teams`' lazy backfill would revoke its ReBAC `public` relation, silently removing an **existing** team from the marketplace as a side effect of a storage repair. Now states `visibility="public"` explicitly (same fidelity rule as `importer.py`), with a test that genuinely fails on revert (pre-flush the attribute reads `None`, not `"private"`).
- `MarketplaceTeams` filter: `visibility === "public"` instead of `!== "private"`. With private as the default, an absent field (version-skew backend) must err on non-disclosure - the old form contradicted the `?? "private"` fallback in the settings panel. New test + fixtures made explicit.
- Docs: annotated the two superseded "default = PUBLIC" statements in place (contract SS30, RFC SS5.1.2) instead of leaving the contract asserting both defaults. SS44 also gained the `create_team` CAN_READ corollary and the one deliberate import exception (a team provisioned from `users.json` references alone carries no visibility, so it lands private).
- Readability: statement-level `if` instead of a ternary hiding an `await`; the test fake's visibility knob defaults to `PRIVATE` (mirroring the real store) rather than a `None` sentinel.

**Filed as follow-up, not bundled** - #2439: with private as the default the private partition is never empty, so `_list_teams` now issues two identical bulk OpenFGA public-tuple reads per listing instead of one (the empty-list short-circuit no longer fires). Fixing it restructures the fred-core rebac-engine helper surface, so it gets its own PR.

**Considered and not acted on**: a single `DEFAULT_TEAM_VISIBILITY` constant to replace the five hand-copied literals, and dropping the response-schema default so codegen makes the TS field required (which would delete the frontend fallback instead of mirroring it). Both are real, both are structural changes beyond this issue's scope.

**Verification after the fixes**: control-plane 851 passed, knowledge-flow 947 passed, fred-core offline 521 passed, frontend 1419 passed, root `make code-quality` green (incl. basedpyright on both touched backends).

simcariou:
## Correction to the review pass - 6a54f2ee5

The `visibility="public"` I added to `backfill_storage_usage.py` in d239baab4 is reverted. Forcing public there was a guess: the script knows a team owns content, never what discoverability its admin intended, and publishing a team on a guess is what the new default exists to prevent. The backfill now states nothing and takes the platform default, same stance as the importer's `users.json`-only provisioning path.

The consequence is stated in SS44 rather than engineered around: a team the backfill materializes that *was* marketplace-listed loses that listing on the next `GET /teams`, and a team_admin re-publishes it deliberately.

`importer.py`'s `row.get("visibility", "public")` stays as-is - for a bundle exported before the field existed, public is not a guess but the exporting platform's actual behavior.

Verification after the revert: knowledge-flow 946 passed, code-quality green.

PR #2444, opened 2026-08-26
fix(#2364): tabular parquet reads on GCS bypass signBlob via server-side ADC downloads

### LINKED ISSUE #2364: [BUG] Tabular parquet reads on GCS (tp-s3ns) fail: bypass signBlob/httpfs for internal-only DuckDB access
## Context

Prod incident on tp-s3ns (Prism, S3NS sovereign cloud): `read_query` and `search_tabular_values` fail for Excel/CSV-backed tabular datasets.

```
"detail": "Error calling the IAM signBytes API: b'{...\"code\": 403, \"message\": \"Permission \\\\'iam.serviceAccounts.signBlob\\\\' denied on resource (or it may not exist).\"...}'"
```

## Root cause

`TabularService._resolve_dataset_location` (`apps/knowledge-flow-backend/knowledge_flow_backend/features/tabular/service.py`) calls `GcsContentStore.get_presigned_url_internal()` so DuckDB's `httpfs` extension can read the Parquet artifact over HTTPS. That method mints a V4 signed URL via IAM `signBlob`, impersonating `signing_service_account_email`.

Per #2318, tp-s3ns does not currently hold `iam.serviceAccounts.signBlob` on that signing service account (GCP-side grant pending) — the same signing SA is shared with control-plane's browser-facing avatars/banners, which are broken for the same reason.

(Ruled out during implementation: a suspected secondary bug — `generate_signed_url()` allegedly signing against the public `googleapis.com` IAM endpoint instead of the sovereign `s3nsapis.fr` one. Checked against the installed `google-cloud-storage==3.1.0`: `Blob.generate_signed_url()` already derives `universe_domain` from `client.universe_domain` internally — i.e. from the same ADC-derived value `build_gcs_client()` mirrors into the client. That path was already correct; the missing IAM grant is the whole story.)

## Why this path doesn't need a signed URL at all

Unlike the browser-facing case (#2318 — media/avatars, which genuinely need a URL an anonymous `` can load), tabular Parquet reads are 100% backend-internal: DuckDB runs in the same process as the code minting the URL. The only reason a URL exists here is that DuckDB's `httpfs` extension wants an HTTP(S) location. Nothing requires `httpfs`/`signBlob` for this path — `GcsContentStore`'s ADC/Workload-Identity-authenticated `storage.Client` already works today for downloads (`get_content`, `get_media`, `get_local_copy`, and `health_check`'s `list_blobs` against both buckets, including the `-objects` bucket where Parquet artifacts live — the pod's readiness proves `storage.objects.list` already works there).

## Fix (implemented)

For GCS, bypass `httpfs`/signed URLs for internal tabular reads entirely: download the Parquet object server-side via the already-working `storage.Client` into a local temp file (mirrors the existing `FileSystemContentStore` fallback in `_resolve_local_dataset_path`), and point DuckDB at that local path instead of a signed URL. No new IAM grant needed beyond plain `storage.objects.get` on the objects bucket (a much lower bar than `signBlob`, and one `list` already proves is at least partially granted).

- `gcs_content_store.py`: new `download_object_to_path()` — same ADC-authenticated blob download already used by `get_content`/`get_media`.
- `tabular/service.py`: `_resolve_dataset_location` downloads to a per-job temp dir for `GcsContentStore` instead of calling `get_presigned_url_internal`; temp dir is scoped to and cleaned up with each DuckDB job (`query_read`, `search_values`, `_load_dataset_frame`).

Out of scope: browser-facing presigned URLs (avatars/media) — that's #2318's proxy-URL design, unaffected by this change.

## Verification

Code review + unit tests only from this side (no live test against tp-s3ns). Live verification — does `storage.objects.get` work on the `-objects` bucket — happens once deployed, coordinated with Sébastien/Arthur. If it also fails, the documented fallback is deploying SeaweedFS instead of GCS for this environment.

florian-muller:

**Issue:** Closes #2364

**Problem in one sentence:** On tp-s3ns (S3NS sovereign cloud), `read_query` and `search_tabular_values` fail with `Permission 'iam.serviceAccounts.signBlob' denied` because tabular Parquet reads mint a V4 signed URL just so DuckDB's `httpfs` extension has an HTTPS location — a signing grant that deployment does not hold.

**Backlog ref:** n/a — GitHub issue is the tracking unit (milestone `swift-v2.2`, Priority: Critical).

---

**RFC written?** Not required — this is a fix that *removes* a dependency rather than adding a design. The change is scoped to swapping one access strategy for another already-working one inside `TabularService`; the durable what/why went straight into the compact doc (`docs/swift/design/DESIGN.md §3`) per the "settled decision → compact doc, not RFC" rule.

**Plan presented to the team before implementation?** Yes — the root-cause analysis and the proposed fix were written up in #2364 itself (issue body: "Why this path doesn't need a signed URL at all" / "Fix (implemented)") before the code landed, including the explicitly ruled-out alternative (the suspected `universe_domain` bug, which turned out to be a non-issue on `google-cloud-storage==3.1.0`).

**Confirmation received?** Issue is assigned and prioritized Critical against the prod incident; live verification on tp-s3ns is coordinated with Sébastien/Arthur once deployed.

---

## 3. What you built

Tabular reads on GCS no longer go through signed URLs or `httpfs` at all. `TabularService` gained `_dataset_location_scope()`, a per-job context manager that yields a resolver turning a `TabularArtifactV1` into a DuckDB-readable location: for `GcsContentStore` it downloads the Parquet artifact server-side into a job-scoped temp directory via the new `GcsContentStore.download_object_to_path()` (the same ADC / Workload-Identity `storage.Client` already used by `get_content`/`get_media`, needing only `storage.objects.get`), and DuckDB reads a local file; every other store keeps its existing behaviour (S3-compatible → internal presigned URL + `httpfs`, filesystem → direct path). The scope wraps the entire DuckDB connection lifetime in all three job sites (`query_read`, `search_values`, `_load_dataset_frame`) because DuckDB opens Parquet lazily at execution time, so temp files must outlive view creation and are cleaned up when the job returns — on success, error, abort and timeout alike.

Three supporting changes came with it:

- **Disk budget.** New `storage.tabular_store.query.max_local_artifact_bytes` (default 512 MiB) with a `_JobDiskBudget` helper: the artifact's *declared* ingestion size is prechecked before a byte is written (so one oversized artifact cannot fill the pod's ephemeral storage mid-download), and actual on-disk bytes are charged after (correct for legacy artifacts with no declared size). Over-budget queries fail as a caller error rather than evicting the pod. `search_values` now mounts lazily one table at a time so its early break on `max_matching_tables` also skips the fetches it would never open.
- **Error redaction.** A missing artifact maps to a generic not-found (never the internal object key); google client errors — whose text embeds the raw GCS REST URL — are re-raised redacted, mirroring the existing `httpfs` signed-URL redaction.
- **Config relaxation.** `content_storage.signing_service_account_email` is now optional: the startup fail-fast in `application_context.py` is deleted, and `get_presigned_url_internal` (still implemented, now unused by any knowledge-flow feature) raises `NotImplementedError` instead of `RuntimeError` when unconfigured, so a capability-probing caller degrades to its unsupported-operation path instead of a 500.

Out of scope, unchanged: browser-facing presigned URLs for avatars/media — that's #2318's proxy-URL design.

---

## 4. Proof of quality

**Tests added or updated:**

| `test_download_object_to_path_roundtrip` | `tests/stores/content/test_gcs_content_store.py` | New store method streams an object to a local file |
| `test_download_object_to_path_missing_raises_and_leaves_no_file` | `tests/stores/content/test_gcs_content_store.py` | 404 raises `FileNotFoundError` and leaves no truncated file behind |
| `test_gcs_content_store_factory_builds_without_signing_email` | `tests/core/test_content_store_factory_gcs.py` | Startup no longer fails when `signing_service_account_email` is unset (replaces the old fail-fast test) |
| `test_tabular_service_reads_gcs_datasets_via_job_local_download` | `tests/services/test_tabular_service.py` | GCS path downloads to a job-local file and never calls `get_presigned_url_internal` |
| `test_tabular_service_gcs_disk_budget_refuses_before_downloading` | `tests/services/test_tabular_service.py` | Declared-size precheck rejects an over-budget artifact with zero bytes downloaded |
| `test_job_disk_budget_charges_actual_bytes_when_no_size_declared` | `tests/services/test_tabular_service.py` | Legacy artifacts with no declared size are charged post-download |
| `test_tabular_service_gcs_missing_artifact_does_not_leak_object_key` | `tests/services/test_tabular_service.py` | Not-found message contains no internal object key |
| `test_tabular_service_gcs_download_errors_are_redacted` | `tests/services/test_tabular_service.py` | GCS client errors surface without the raw REST URL |
| `test_tabular_service_gcs_search_downloads_lazily_until_match_cap` | `tests/services/test_tabular_service.py` | `search_values` stops downloading once the match cap is hit |

```
************ All code quality checks completed ************
```

**Raw `basedpyright` output:**

```
0 errors, 0 warnings, 0 notes
```

**`make test` output (touched suites):**

```
55 passed, 3 warnings in 5.49s
```

(`tests/services/test_tabular_service.py tests/stores/content/test_gcs_content_store.py tests/core/test_content_store_factory_gcs.py`)

---

## 5. Docs updated

| ----------------------------------------------------------------- | ------------ |
| Backlog `[ ]` item is now done | n/a — backlog files are frozen; GitHub issue #2364 is the tracking unit |
| New behaviour, API field, or contract change | `docs/swift/design/DESIGN.md` §2 (Locate step) and §3, rewritten as "Tabular Reads on GCS — Server-Side Downloads (#2364)" |
| Frozen contract touched (`execution.py`, `agent_app.py`, OpenAPI) | n/a — no runtime/control-plane contract surface touched |
| UX component implemented or visual status changed | n/a — backend only |
| WORKPLAN sprint item finished | n/a — frozen |
| Code and a design doc now diverge | Fixed in the same change (DESIGN.md §3 rewritten); config descriptions updated in `common/structures.py`, regenerated `config/schema/configuration.schema.json` + `deploy/charts/fred/values.schema.json`, and `deploy/charts/fred/values-gcp.yaml` comments/IAM guidance |

---

## 6. Close-out statement

```
## Task close-out
- Code: GCS tabular Parquet reads download artifacts server-side via the ADC storage.Client into a job-scoped temp file (new GcsContentStore.download_object_to_path + TabularService._dataset_location_scope) instead of minting V4 signed URLs for httpfs; signing_service_account_email becomes optional and a per-job disk budget (max_local_artifact_bytes, 512 MiB) bounds local materialization.
- Tests: 55 passed across the three touched suites; 9 tests added/replaced covering the download path, the disk budget (pre- and post-check), error redaction, lazy search mounting, and the relaxed factory.
- Docs updated: docs/swift/design/DESIGN.md, apps/knowledge-flow-backend/knowledge_flow_backend/common/structures.py, apps/knowledge-flow-backend/config/schema/configuration.schema.json, deploy/charts/fred/values.schema.json, deploy/charts/fred/values-gcp.yaml
- Tracking: closes #2364
- Skipped steps: Step 1 (RFC) — settled fix that removes an IAM dependency, not an open design question; the what/why went into DESIGN.md per the RFC-vs-doc rule.
```

---

## 7. Risk and rollback

**What breaks if this is wrong?** The blast radius is the GCS content-store path only — MinIO/S3 and filesystem deployments take the unchanged branches. Two residual risks: (1) if the deployment lacks `storage.objects.get` on the `-objects` bucket the reads still fail (the pod's readiness already proves `storage.objects.list` works there, so this is unlikely; documented fallback per #2364 is deploying SeaweedFS for that environment); (2) whole-artifact downloads replace `httpfs` range reads, so a very large Parquet costs more latency and pod-local disk than before — bounded by `max_local_artifact_bytes`, and a pod-local cache keyed by object key + etag is the natural follow-up if latency warrants it. Live verification against tp-s3ns has not happened yet — this is code review plus unit tests from this side.

**How do we roll back?** Revert the single commit. Redeploying then also requires re-setting `content_storage.signing_service_account_email` in values (the old code fails fast at startup without it) — that's the only config coupling.

### DIFF
--- a/apps/knowledge-flow-backend/config/schema/configuration.schema.json
+++ b/apps/knowledge-flow-backend/config/schema/configuration.schema.json
@@ -416,7 +416,7 @@
             }
           ],
           "default": null,
-          "description": "Service account email used to sign V4 signed URLs for backend-internal tabular Parquet reads, via IAM signBlob under Workload Identity (no JSON key). Required for content_storage.type=gcs; startup fails clearly when omitted. The Workload Identity service account must hold iam.serviceAccounts.signBlob on this account, which must have storage.objects.get on the objects bucket.",
+          "description": "Optional service account email used to sign V4 signed URLs via IAM signBlob under Workload Identity (no JSON key). Tabular Parquet reads no longer need it: they download artifacts server-side with the ADC client, which only requires storage.objects.get on the objects bucket (#2364). Set it only to re-enable internal V4 signed URLs for code paths that explicitly call get_presigned_url_internal.",
           "title": "Signing Service Account Email"
         }
       },
@@ -2370,13 +2370,13 @@
         "access_mode": {
           "const": "presigned_url",
           "default": "presigned_url",
-          "description": "Primary object-access method for remote tabular artifacts.",
+          "description": "Primary object-access method for remote tabular artifacts on S3-compatible stores. GCS ignores it: artifacts are downloaded server-side to a job-local file instead (#2364).",
           "title": "Access Mode",
           "type": "string"
         },
         "internal_presigned_ttl_seconds": {
           "default": 3600,
-          "description": "TTL in seconds for backend-internal object-storage URLs used by tabular DuckDB reads.",
+          "description": "TTL in seconds for backend-internal object-storage URLs used by tabular DuckDB reads on S3-compatible stores. Unused on GCS (no URLs are minted).",
           "minimum": 1,
           "title": "Internal Presigned Ttl Seconds",
           "type": "integer"
@@ -2435,6 +2435,13 @@
           "description": "Maximum datasets one request may mount or scan. Backstop for searches and for queries whose referenced-relation set is still too large.",
           "minimum": 1,
           "title": "Max Selected Datasets",
+          "type": "integer"
+        },
+        "max_local_artifact_bytes": {
+          "default": 536870912,
+          "description": "Per-job disk budget, in bytes, for tabular artifacts materialized onto the pod's local storage (GCS access path, #2364). A query needing more fails as a caller error (400) instead of filling the pod's ephemeral storage; the declared artifact size is checked before downloading. This is also the effective per-artifact ceiling on GCS - raise it on deployments that ingest very large tabular files. Instantaneous worst case per pod: max_concurrent_queries times this value.",
+          "minimum": 1,
+          "title": "Max Local Artifact Bytes",
           "type": "integer"
         }
       },
--- a/apps/knowledge-flow-backend/knowledge_flow_backend/application_context.py
+++ b/apps/knowledge-flow-backend/knowledge_flow_backend/application_context.py
@@ -558,18 +558,12 @@ class ApplicationContext:
                 public_secure=config.public_secure,
             )
         elif isinstance(config, GcsStorageConfig):
-            # Fail fast at startup: GCS tabular Parquet reads need a signing SA to
-            # mint V4 signed URLs via IAM signBlob. There is no per-feature flag to
-            # detect tabular usage, so a missing email must stop the boot with a
-            # clear message rather than surfacing later as an opaque runtime error.
-            if not config.signing_service_account_email:
-                raise ValueError(
-                    "content_storage.type=gcs requires 'signing_service_account_email' "
-                    "to sign V4 signed URLs for tabular Parquet reads (IAM signBlob under "
-                    "Workload Identity, no JSON key). Set it to the signing service account "
-                    "that holds storage.objects.get on the objects bucket and on which the "
-                    "Workload Identity service account has iam.serviceAccounts.signBlob."
-                )
+            # signing_service_account_email is optional: tabular Parquet reads
+            # download artifacts through the ADC/Workload-Identity client instead
+            # of minting V4 signed URLs (#2364 — signBlob is not grantable on
+            # every deployment), and browser-facing presigned URLs are
+            # unsupported on GCS by design. The email is consumed only if a code
+            # path explicitly calls get_presigned_url_internal.
             self._content_store_instance = GcsContentStore(
                 document_bucket=f"{config.bucket_name}-documents",
                 object_bucket=f"{config.bucket_name}-objects",
--- a/apps/knowledge-flow-backend/knowledge_flow_backend/common/structures.py
+++ b/apps/knowledge-flow-backend/knowledge_flow_backend/common/structures.py
@@ -196,12 +196,12 @@ class GcsStorageConfig(BaseModel):
     signing_service_account_email: Optional[str] = Field(
         default=None,
         description=(
-            "Service account email used to sign V4 signed URLs for backend-internal "
-            "tabular Parquet reads, via IAM signBlob under Workload Identity (no JSON "
-            "key). Required for content_storage.type=gcs; startup fails clearly when "
-            "omitted. The Workload Identity service account must hold "
-            "iam.serviceAccounts.signBlob on this account, which must have "
-            "storage.objects.get on the objects bucket."
+            "Optional service account email used to sign V4 signed URLs via IAM "
+            "signBlob under Workload Identity (no JSON key). Tabular Parquet reads "
+            "no longer need it: they download artifacts server-side with the ADC "
+            "client, which only requires storage.objects.get on the objects bucket "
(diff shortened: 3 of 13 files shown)


florian-muller:
The two RFC commits (ac688b2a3, 76674830c) went into `swift` as 778070737 and 6b0d16307. The GCS signBlob code fix (28141b308) this branch was based on is left out, so I'm closing this PR.

Issue #2456, opened 2026-08-27
docs(chart): the migration defaults carry a stale comment


simcariou:
`deploy/charts/fred/values.yaml` ships both backends with a stale comment:

```yaml
  knowledge-flow-backend:
    migration:
      enabled: false    # Enable when knowledge-flow-backend gets Alembic migrations
  control-plane-backend:
    migration:
      enabled: false    # Enable when control-plane-backend gets Alembic migrations
```

Each backend has owned an Alembic tree for months (`alembic_version_knowledge_flow`,
`alembic_version_control_plane`), so both comments read as "this backend has no
migrations yet" and are simply false.

The defaults themselves stay off on purpose: a platform overlay either turns the Job
on in its own values or runs the upgrade imperatively. Only the comments are wrong.

Worth stating plainly in them, since #2314: knowledge-flow creates no table at startup
any more (`require_tables(REQUIRED_TABLES)`), so a deployment that leaves the Job off
and never migrates gets a `CrashLoopBackOff` with
`[SCHEMA] knowledge-flow: required table(s) missing`. There is a cross-tree dependency
too - `REQUIRED_TABLES` includes `users` and `teammetadata`, migrated by the
control-plane tree.

**Scope:** comments only, defaults unchanged.

simcariou:
Landed on swift as fc951247e: comments rewritten on all three blocks, and the fred-agents default joins the other two at `enabled: false` - whether the Job runs is the platform overlay's call, not the chart's.

Issue #2314, opened 2026-08-10
Alembic table-ownership filter is inert: _owned_tables derived from shared CoreBase admits every backend's tables


florian-muller:
## Problem

`fred_core.models.base.Base` (`CoreBase`) is a single shared declarative base with **no ownership boundary**. Tables registered on it are migrated by different Alembic trees, but nothing in the code expresses or enforces which tree owns what:

| Table on `CoreBase` | Migrated by | Version table |
|---|---|---|
| `tag`, `metadata` | knowledge-flow tree | `alembic_version_knowledge_flow` |
| `users`, `session`, `session_metadata`, `teammetadata`, `task_run`, `task_event_log` | control-plane tree | `alembic_version_control_plane` |
| `session_history` | fred-runtime tree | (fred-runtime) |

Both apps' Alembic environments pass **both** metadata objects to autogenerate:

- `apps/knowledge-flow-backend/alembic/env.py:42` — `target_metadata=[Base.metadata, CoreBase.metadata]`
- `apps/control-plane-backend/alembic/env.py:37` — `target_metadata=[Base.metadata, CoreBase.metadata]`

So **each tree's autogenerate sees the other tree's tables**, with no filter. Control-plane's env even imports `fred_core.documents.document_models` ("registers metadata table with CoreBase") while no control-plane migration ever creates `tag` or `metadata` — those belong to knowledge-flow.

Note this is *not* a case of misplaced models: `tag` and `metadata` are genuinely shared. Control-plane's import/export subsystem reads and writes both directly (`import_export/exporter.py:30-31`, `importer.py:93-95`, `teardown.py:47-48`, `api.py:39-40`), plus the `documents_total` KPI preset. Living on `CoreBase` is correct — what is missing is per-tree ownership scoping.

## Impact

- **Autogenerate proposes foreign tables.** Running `alembic revision --autogenerate` in control-plane against a database lacking `tag` / `metadata` produces a migration that creates them — a third writer for tables knowledge-flow already owns. The same holds in reverse for knowledge-flow and `users` / `teammetadata`.
- **Duplicated migrations already exist.** The task_run acknowledgement columns were added twice, independently, in two trees: `d96d9ae21375` (KFB) and `9e5074103b67` (CP). Both were written defensively (inspect-before-add), which is evidence the collision was already hit and patched at the symptom level rather than the cause.
[...]

PR #2457, opened 2026-08-27
fix(#2456): run the control-plane and knowledge-flow migration jobs by default

simcariou:
**Rolled back - this change is not the one we want.** The chart's migration
defaults stay `false` on purpose: a platform overlay either turns the Job on in
its own values or runs the upgrade imperatively. Only the stale comments needed
fixing, which is now the whole scope of #2456.

Original description below, kept for the record.

---

Since #2314 knowledge-flow creates no table at startup - both entrypoints call
`require_tables(REQUIRED_TABLES)` and refuse to boot on an unmigrated database.
The chart's `knowledge-flow-backend` and `control-plane-backend` values still
carried `migration.enabled: false` commented "Enable when <backend> gets Alembic
migrations", which has been false for months.

This PR flipped both defaults to `true`. That part is reverted; the comment fix
lands separately.

### DIFF
--- a/deploy/charts/fred/values.yaml
+++ b/deploy/charts/fred/values.yaml
@@ -1096,9 +1096,12 @@ applications:
     job:
       enabled: false
     migration:
-      enabled: false    # Enable when knowledge-flow-backend gets Alembic migrations
-      scaleDown: true
-      backoffLimit: 3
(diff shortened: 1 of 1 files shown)


simcariou:
Rolled back before the release: the defaults stay off by design, only the stale comments are fixed. Tracked back on #2456.

Issue #1772, opened 2026-06-18
feat(knowledge-flow): targeted similarity / comparison search (KF-SIMILARITY-SEARCH)


dimitri-tombroff:
## Summary
Add a **targeted similarity search** mode to swift knowledge-flow: given an anchor
text, return the passages most similar to it **within a set of targets named on the
call** (documents and/or folders), **ranked best-first**.

Additive — sits alongside the existing conversational/question-answering search,
which is unchanged.

**RFC:** `docs/swift/rfc/KNOWLEDGE-FLOW-SIMILARITY-SEARCH-RFC.md` (KF-SIMILARITY-SEARCH)

## Why
Some agents aren't doing Q&A — they do **document-to-document comparison**. The rags
assessment agent (Eva) compares an information system's technical-architecture doc
against its operation manual to derive similarities, contradictions, and a
technical-detail sheet, then cross-checks against the CMDB. Every step is the same
primitive: *"for **this** passage, find the closest passage **in that** document"* —
re-aimed **per query**. Today's search fixes its document scope **once per
conversation**, so this can only be faked via fetch-everything-then-filter.

## Scope
- swift `apps/knowledge-flow-backend` only.
- **MCP surface first** (agents/Eva); REST is a follow-up.
- Additive: no change to existing conversational search; no ingestion/embedding/
  storage changes (reuses the already-indexed corpus); same auth model.

## Functional contract
**Inputs:** `anchor` text · `targets` (document and/or folder ids — empty = error) ·
`top_k` · `rerank` (default on) · optional `min_score`.
**Output:** matches **ordered best-first**, each with matched passage text, source
(document id + name, location) and a relevance score. Results drawn **only** from the
named targets.

"Compare A to B" = for each passage of A, call with `anchor=passage`, `targets=[B]`,
`top_k=1`.

## Acceptance criteria
- [ ] One call returns passages most similar to an anchor, **restricted to named
      documents/folders**, ranked best-first with source + score.
- [ ] Targeting honoured — never returns passages outside the named targets.
- [ ] Re-ranking on by default.
- [ ] Existing conversational search unchanged (no regression).
- [ ] Exposed over **MCP**; REST tracked as a follow-up.
- [ ] Eva can drop its client-side fetch-then-filter workaround and call this directly.

## Decisions to confirm with the rags team (RFC §8)
- [ ] Targeting granularity: documents + folders for v1 (defer passage-level)?
- [ ] Rerank on by default?
[...]

PR #2463, opened 2026-08-28
feat(#2461): document_similarity capability - targeted document-to-document comparison

### LINKED ISSUE #2461: feat(capabilities): document_similarity - targeted document-to-document comparison
## Context

`knowledge-flow` already ships targeted similarity / comparison search (#1772,
KF-SIMILARITY-SEARCH): `SimilaritySearchRequest`, `POST /vector/similarity-search`,
the service orchestration and the pgvector / in-memory store support are all on
`swift`, documented in `docs/swift/design/DESIGN.md` §4.

What was missing is the consumer side: a Fred agent had no first-order way to call
it. The endpoint is reachable over the Text MCP server, but not as a capability an
agent's team can enable, configure and scope.

That gap is what still blocks #1772's last acceptance criterion ("Eva can drop its
client-side fetch-then-filter workaround and call this directly").

## Scope

Expose similarity search as a **first-order capability**, `document_similarity`:

- a capability under `fred_runtime/capabilities/`, registered through the
  `fred.capabilities` entry point next to its `document_access` /
  `document_summarize` / `document_verbatim` / `document_extract` siblings,
  `ADMIN_GATED` like them, with one tool `find_similar_passages` and per-instance
  config for match count, rerank and relevance floor
- a new typed port `RuntimeServices.document_similarity`
  (`DocumentSimilarityPort.find_similar`) with its runtime adapter - deliberately
  not a second method on `DocumentSearchPort`, whose targeting comes from stored
  config rather than from the model on each call
- `VectorSearchClient.similarity_search` for the wire call, with a read timeout
  sized for the cross-encoder pool and a transient retry scoped to failures where
  no work is in flight
- the frontend Simple-view pack that groups the uid-targeted document capabilities

## Not in scope

The `mvp/rags-support` branch carries this alongside a lot of unrelated work. Only
the slice above is ported. Explicitly left there: `supports_capabilities` on
`AgentDefinition` / `GraphAgentDefinition` (the graph-capability bridge),
`FakeGraphNodeContext.invoke_tool`, the `comparison` / Eva agent,
`InformationSystemsPage`, the `rags` frontend slice, the checkpoint-namespace
fixes, and that branch's Langfuse / `trace_payloads` state (it is behind `swift`).

That branch also exposes this as a `knowledge.similarity_search` built-in tool ref.
That shape is deliberately NOT what ships here: `document_access` documents the
built-in surface as back-compat whose retirement is a follow-up, so adding to it
would mean shipping onto a surface with a scheduled end, and a second,
differently-scoped comparison path the moment anyone wires both.

simcariou:
Closes #2461.

Brings targeted document-to-document comparison to Fred agents as a **first-order
capability**, `document_similarity`.

## What was already here

Knowledge Flow's side shipped with #1772 and needed nothing: `SimilaritySearchRequest`,
`POST /vector/similarity-search`, the service orchestration, the pgvector and in-memory
store support, its contract test, and `DESIGN.md` §4. What was missing was a caller -
the endpoint was reachable over the Text MCP server but had no first-order path, which
is what still blocked #1772's last acceptance criterion.

## What this adds

- **`document_similarity` capability** (`libs/fred-runtime/fred_runtime/capabilities/`),
  registered through the `fred.capabilities` entry point next to `document_access` /
  `document_summarize` / `document_verbatim` / `document_extract`, `ADMIN_GATED` like
  them. One tool: `find_similar_passages(anchor, document_uids, top_k)`. Per-instance
  config for match count, rerank, and relevance floor.
- **`RuntimeServices.document_similarity`**, a new typed port
  (`DocumentSimilarityPort.find_similar`) behind `DocumentSimilarityAdapter`.
- **`VectorSearchClient.similarity_search`** for the wire call, with its own read
  timeout and a tightly-scoped transient retry.
- **Frontend**: the Simple-view pack is retitled "Working with documents" /
  "Travail sur documents" and now carries all three uid-targeted document
  capabilities. Corpus-wide discovery stays in the team-resources pack.

## Two design notes

**Why a capability, not a built-in tool ref.** The first commit on this branch ported
the `mvp/rags-support` shape verbatim - a `knowledge.similarity_search` entry in the
`fred-sdk` built-in catalog. That is removed here, not kept alongside.
`capabilities/document_access/capability.py` already documents the built-in surface as
back-compat whose retirement is a follow-up, so adding to it would have meant shipping a
new tool onto a surface with a scheduled end, and a second, differently-scoped comparison
path the moment anyone wired both. The capability path also buys per-instance config,
admin gating and team scoping.

**Why its own port, not a method on `DocumentSearchPort`.** Adding an abstract method to
a published SDK ABC breaks every out-of-tree implementor, and every other document
feature here already has its own optional port. More importantly the two enforce
different things: under `search(...)` the uids come from the capability's stored config,
so narrowing them against the session binding is a formality; under `find_similar(...)`
they come from the **model**, on the call, so that seam is the only thing between an
LLM-named uid and a document the user never put in this conversation.

## Fixed along the way

Each of these has a test:

- **The pack switch would have been stuck off for every existing team.**
  `derivePackChecked` required *every* id in a pack to be selected, while
  `applyPackToggle` only ever adds admin-enabled ones - so a pack containing anything
  unavailable could not be switched on at all (flip it, the missing member is never
  added, the derived state reads false again), while switching it off still stripped the
  rest. Latent before this branch (an admin enabling only part of a pack), but adding an
  `ADMIN_GATED` capability to an existing pack would have hit everyone on upgrade.
  `derivePackChecked` now takes `availableIds`.
- **A scope refusal read as "no matches".** When narrowing against the session binding
  emptied the target set, the adapter returned no hits - indistinguishable, to the model,
  from a genuine no-match, so it would report "nothing in that document resembles this
  passage" about a document it never searched. It now raises `DocumentScopeRefusedError`,
  which the tool renders as its own error naming the out-of-scope uids.
- **The docstring pointed at attachments.** It told the model to resolve uids from the
  conversation's attached files, but KF runs this mode with `include_session_scope=False`
  - those uids can only ever return nothing.
- **Pool timeouts were not retried** on the rationale that work may be in flight. A pool
  timeout means no connection was ever acquired, so nothing was sent.
- **Every 5xx was retried**, including the plain 500 KF wraps unexpected exceptions in -
  three full pool-and-rerank round trips per anchor passage before failing anyway. Now
  502/503/504 only.
- **The heaviest read ran on the shared 30s default.** KF retrieves a pool of up to 100
  chunks and cross-encodes all of them in one request, and a read timeout is the one
  failure the retry deliberately will not recover from. It gets `similarity_read`
  (default 90s), mirroring the `summarize_read` precedent.

## Verification

- `make code-quality` from the monorepo root: clean
- `fred-runtime` 1039 passed, `fred-sdk` 338 passed / 3 skipped, `fred-agents` 46 passed
- frontend: 30 passed across the agent-form suites
- 33 new/updated tests across five files
- `/code-review` at high effort, twice (once per approach); every finding applied
- config schemas regenerated via `make generate-config-schema`, not hand-edited


simcariou (commented):
Tester le reranking avant merge !!

simcariou:
reranking ok à tester en offline lundi

PR #2466, opened 2026-08-28
docs(#2465): cap code comments at 2-3 lines

### LINKED ISSUE #2465: docs(CLAUDE.md): cap code comments at 2-3 lines
Comments written by the assistant keep growing into paragraphs that re-narrate the decision, list the alternatives considered, and restate what the next line does. That is noise the next reader has to skim past, and it goes stale faster than the code around it.

`docs/CONVENTIONS.md` covers code style for humans; `CLAUDE.md` is where the assistant's operational rules live, so the cap belongs next to Step 4 of the task lifecycle: say the "why" the code cannot show, then stop. When the reasoning genuinely needs more room, it goes in the relevant compact doc with a pointer from the code.

simcariou:
Closes #2465.

Adds one rule next to Step 4 of the task lifecycle in `CLAUDE.md`: say the "why" the code cannot show, then stop. A comment that re-narrates the decision, lists the alternatives, or restates the next line is noise the next reader skims past, and it goes stale faster than the code. Long reasoning goes in the relevant compact doc with a pointer from the code; file headers get a few lines on what the module is for, not an essay.

Instruction-only change - no code touched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,6 +180,14 @@ implement authorless, untracked work.
 
 **Step 4 — Implementation.** Write the code. Coding constraints: `docs/CONVENTIONS.md`.
 
+**Keep code comments to 2-3 lines.** Say the "why" the code cannot show, then
+stop. A comment that re-narrates the decision, lists the alternatives, or
+restates what the next line does is noise the next reader has to skim past —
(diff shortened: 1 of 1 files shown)


simcariou:
Folded into #2464 - the rule lands with the change that motivated it.

Issue #2458, opened 2026-08-27
feat(OPSCAP-01-PG): platform_postgres capability — read-only SQL over the platform DB + platform_ops template


florian-muller:
First capability of the admin-ops family (OPSCAP-01-PG). Spec validated 2026-08-27:
`docs/swift/rfc/admin-ops-capabilities/PLATFORM-POSTGRES.md` (parent RFC
`docs/swift/rfc/ADMIN-OPS-AGENTS-RFC.md`).

Read-only SQL over the platform Postgres database, Tier B (reuses the pod's
`storage.postgres` credentials through a new generic `PlatformSqlPort` runtime port).
Proves the whole admin-ops chain: package → entry point → admission → grant → enroll →
chat, and ships the first ready-made template.

## Scope (one PR)

- **fred-sdk**: `PlatformSqlPort` protocol + `SqlQueryResult` + `PlatformSqlPortError` in `fred_sdk/contracts/runtime.py`.
- **fred-runtime**: adapter beside the engine wiring — dedicated pool (size 2, no overflow) from `storage.postgres`, server-side read-only enforcement (single statement by construction via extended protocol, explicit `READ ONLY` transaction, `default_transaction_read_only=on`), 200-row cap, clamped statement timeout; `RuntimeServices.platform_sql` wiring.
- **`libs/fred-capability-platform-ops/`**: in-tree package scaffold + `postgres/` capability — tools `postgres_list_tables` + `postgres_run_query(sql)`, one config field `statement_timeout_s` (default 15, 1–120), JSON result envelope with structure-preserving truncation (200 rows / ~1000 chars per cell / ~40 KB total).
- **fred-agents**: `platform_ops` ready-made template (default `["platform_postgres"]`) + prompt file; en/fr i18n keys.
- **Docs**: capability row per AUTHORING.md.
- **Tests**: capability unit tests (stubbed port, truncation edges, boot invariant); adapter integration tests gated behind a Postgres DSN proving the three §3 rejections (multi-statement, write in transaction, `SET … read_only = off` escape).

No frozen-contract change expected (additive capability + template). WP1 (#1974 default-capability selection) has landed, so the zero-click template flow is unblocked.

florian-muller:
[...]

Issue #2175, opened 2026-07-30
REASON-01 Amendment B: let an agent author preselect reasoning ON for new conversations


Timothe-Le-Chatelier:
Design: `docs/swift/rfc/MODEL-REASONING-ENABLEMENT-RFC.md` §16 (Amendment B). Follow-on to #2166 (closed).

## Problem

REASON-01 level 3 gives an agent author one binary: offer the composer's reasoning toggle, or don't. The value that toggle **starts at** is hardcoded — `params={"default": False}` in `_platform_reasoning_control`.

For agents whose purpose is deliberation, every user must find and flip the switch in every new conversation. Most won't: the feature ends up offered and unused.

## Change

One field, `reasoning_default_on`, on the same three surfaces as `reasoning_enabled` (`ManagedAgentTuning`, create/update requests, instance summary). It seeds `ChatControlDescriptor.params.default` for the `reasoning_toggle` widget — the seam `useComposerSettings` already reads (RFC §7.4). No runtime change, no new channel.

UI: a second switch nested under "Reasoning" in the agent form's General section, rendered only while the offer is on.

## Two properties, both tested

1. **A preselection cannot conjure a control.** §8's gates run first and unchanged — offer off, or no model reasoning platform-wide, still means no descriptor at all. Guards against: author leaves default on, withdraws the offer, reasoning quietly stays on.
2. **The two fields never write each other.** Withdrawing the offer leaves the default stored but inert, so toggling the offer off and back on recovers the author's choice.

## Safety (RFC §9)

§9's mitigation is level 2 (the platform-admin per-model off switch) and is untouched. Default stays `False`, so nothing changes for any agent whose author does not act, and the form hint states the cost — slower, and on tool-using agents can repeat tool calls (`AGENT-THINKING-API-RFC.md` Amendment C) — at the point of decision.

Issue #1974, opened 2026-07-10
Capability selection end-to-end: catalog aggregation, save-time validation, tuning slices


florian-muller:
## Parent

#1961 — Agent Capability RFC (`CAPAB-01`, RFC §3.4, §3.7–§3.9)

## What to build

The product path that lets a user put a capability on an agent, end to end:

- Control-plane aggregates capability manifests from pods into the product catalog, the way runtime templates are aggregated today.
- `ManagedAgentTuning` gains `selected_capability_ids` and `capability_config` — opaque, pod-validated slices stored as `{"schema_version": manifest.version, "config": {...}}` — alongside the existing MCP trio for now (the trio retires in the McpCapability issue).
- Agent save round-trips to the pod: the platform enforces `AssetSlot` cardinality/extension per slot with generic, uniformly-worded 422s **before** capability code runs; the capability's `validate_config(config, uploads, ctx)` owns content validation and returns the stored config, persisted verbatim. Asset binaries go through the KF-backed asset store; only storage keys enter the stored config — blobs never enter `tuning_json`.
- Save-time availability check: `selected_capability_ids` is validated against the capabilities the instance's bound pod advertises; unknown ids → typed 422.
- Agent assembly validates each `capability_config` slice against `StoredConfigModel`, running the optional lazy `upgrade_config` hook on `schema_version` mismatch (never a mass row migration); the typed result is what `CapabilityContext.config` carries.
- Frontend: the Tools tab lists capabilities from the catalog; `config_fields` render through the existing metadata-driven agent-creation form (zero new form code for scalars).

## Acceptance criteria

- [ ] Capability catalog visible in agent creation, aggregated from pod manifests
- [ ] Enabling a capability + filling its config on an agent persists a pod-validated stored config in `tuning_json`
- [ ] Asset-slot cardinality/extension violations rejected with uniform 422s before capability code runs; binaries stored via KF, only keys persisted
- [ ] Unknown capability id at save rejected with typed 422
- [ ] Assembly-time slice validation and the `upgrade_config` path covered by tests
- [ ] Demo capability creatable and usable end-to-end through the UI

## Blocked by

- #1973

dimitri-tombroff:
Closing as delivered. Capability catalog aggregation, selected capability/config slices, save-time validation and runtime assembly are present and documented in the current product/runtime contracts.

Issue #1780, opened 2026-06-19
Reasoning models emit redundant identical tool calls in the Fred agentic loop — root-cause fix + guardrail


dimitri-tombroff:
## Summary

After switching Fred's tool-calling ReAct agent from `mistral-medium` (no reasoning)
to `mistral-small-2603` with `reasoning_effort: high`, the agent emits the **same tool
call with byte-identical arguments 3–5 times in a row** within a single user turn
before answering. The repeats return identical results — pure wasted latency and tokens.

Full root-cause analysis, state-of-the-art survey, and recommended solution:
`docs/swift/issues/ISSUE-005-reasoning-model-redundant-tool-calls.md`

## Root cause (one line)

Fred strips provider-native reasoning blocks between loop steps (a workaround for
Mistral's HTTP 422 on replayed raw reasoning), and `langchain-openai`'s `ChatOpenAI`
can't carry the reasoning field over a `base_url` provider anyway. Without its reasoning
threaded back, the model "wakes up" each step having forgotten why it called the tool,
re-derives the same plan, and re-issues the identical call. `reasoning_effort: high`
amplifies it. This is an industry-wide reasoning-era failure mode, not a Fred/Mistral bug.

## Scope of work (~3–5 engineer-days, front-loaded)

1. **Reproduce + instrument** — confirm reasoning is absent on replay and tool-call args
   are byte-identical across repeats; capture a clean regression trace.
2. **Ship the guardrail (keep permanently)** — `(tool_name, canonical_args)` de-dup +
   per-turn iteration cap in the LangGraph loop. Set `reasoning_effort: medium` and
   `parallel_tool_calls=false` as interim hardening.
3. **Spike Pydantic AI on the Mistral endpoint** — validate `openai_chat_thinking_field`
   + `openai_chat_send_back_thinking_parts` round-trip `reasoning_content` and that the
   422 disappears with structured (not raw-string) reasoning replay.
4. **Integrate as the model-access layer** — swap the model client behind the existing
   LangGraph nodes; keep orchestration untouched. Invert the rule from "always strip"
   to "thread reasoning within the open loop, strip on closed turns."
5. **Regression-test** — single-topic question must produce one `knowledge_search` call,
   then an answer, even at `reasoning_effort: high`.
6. **Fallback** — if the Pydantic AI path is fragile: route via LiteLLM and re-attach
   `thinking_blocks`, or adopt the reasoning-as-tools pattern.

## Acceptance criteria

[...]

PR #2467, opened 2026-08-28
feat(#2458): platform_postgres capability — read-only SQL over the platform DB + platform_ops template

florian-muller:
> **Note on how this PR came to exist:** the five commits below were first pushed
> straight to `swift` (merged as `2da3fd51`), bypassing review. That merge has been
> reverted on `swift` (`517b05fd`) and the identical five commits are re-proposed
> here so the change gets its normal gate. The diff is byte-identical to what was
> briefly on `swift` — and that content already went green through the full push CI
> on `swift` (Code Quality, Alembic, Config Schemas, Docker Build, CodeQL).

**Issue:** #2458 (also lands WP1 of #1974 — see below)

**Problem in one sentence:** Operators have to shell into `psql` to answer routine
questions about their own Fred deployment (teams, agents, sessions, usage), and
there is no capability-safe way for an agent to read the platform database.

**Backlog ref:** n/a — GitHub Issues/Milestones are the tracking surface
(`swift-v2.2`); `BACKLOG.md` is frozen.

#### Metrics

image

#### AI act check

image

#### Template and capability dependency

image

---

**RFC written?**

Yes, two:
- `docs/swift/rfc/ADMIN-OPS-AGENTS-RFC.md` — parent RFC for the admin-ops
  capability family (packaging, security model, template mechanics).
- `docs/swift/rfc/admin-ops-capabilities/PLATFORM-POSTGRES.md` — the per-capability
  spec (OPSCAP-01-PG), validated 2026-08-27.

**Plan presented to the team before implementation?**

Yes — the spec above is the plan: decisions-at-a-glance table, capability surface,
port design, result formatting, template, observability, test strategy, delivery
scope. Written and reviewed before implementation.

**Confirmation received?**

Yes — spec signed off 2026-08-27, tracked on #2458 (milestone `swift-v2.2`).

---

## 3. What you built

**A new generic runtime port.** `PlatformSqlPort.execute_read(sql, *, timeout_s)`
in `fred_sdk/contracts/runtime.py`, with `SqlQueryResult` and the typed
`PlatformSqlPortError`, wired as the optional `RuntimeServices.platform_sql`. The
contract is deliberately backend-agnostic — nothing Postgres-, pool-, or
policy-specific — so it is re-implementable against another backend and reusable by
any future capability wanting read-only platform SQL.

**A credentialed adapter in fred-runtime** (`app/platform_sql.py`) built from the
pod's own `storage.postgres` credentials (Tier B — zero operator setup), on a
dedicated pool (size 2, `max_overflow=0`) so a heavy analytical query can never
starve the checkpointer/history hot path. Read-only is enforced server-side in
three layers, with **no SQL string validator**:

1. **single statement by construction** — the extended query protocol's parse step
   rejects `SELECT 1; DELETE …` before anything runs;
2. an explicit **`READ ONLY` transaction** around every execution;
3. **`default_transaction_read_only = on`** set at connect time.

Layer 1 is the load-bearing one: layers 2 and 3 are transaction/session *state*, and
only a multi-statement script can reach the commands that mutate that state
(`SET default_transaction_read_only = off; COMMIT; DELETE …` is legal inside a READ
ONLY transaction right up until it isn't). A Python-side parser is deliberately
omitted — it is the only layer that could *wrongly reject* legitimate CTEs, EXPLAIN,
or comments, and per the parent RFC it is never the guarantee. Plus a 200-row cap
(fetch 201 to set `row_limit_hit`) and a timeout clamped to [1, 120] applied
server-side.

**The capability package** `libs/fred-capability-platform-ops/`
(`PlatformPostgresCapability`), two tools per the family's minimal-surface principle:
`postgres_list_tables()` for zero-arg grounding and `postgres_run_query(sql)`.
Anything a single query can express (describe-table, EXPLAIN, FK lookup) is *not* a
tool. One config field, `statement_timeout_s`. Because the port is the capability's
only path to the database, "no credentials in capability code" is structural rather
than a code-review promise.

**Result envelope** with truncation applied in order — 200 rows, then ~1000 chars
per cell cut in place with a `…[truncated, N chars]` marker, then a ~40 KB backstop
cut at *row* granularity. Per-cell before total is deliberate: with a total cap only,
one giant cell eats the budget and drops every other row, or the cut lands mid-cell
and breaks the table shape the model is parsing.

**The `platform_ops` template** (`fred.github.platform_ops`) defaulting to
`["platform_postgres"]`, so an operator enrolls an admin-ops agent with zero extra
clicks; later admin-ops capabilities append themselves to this list. Prompt is
editable via a `prompts.system` FieldSpec.

**WP1 (`feat(#1974)`, last commit).** The spec's §5 ordering dependency, which
turned out **not** to be on `swift` after all: a template's declared
`default_capability_ids` never reached a new instance, because the agent form
always submits an explicit `capability_ids` and the backend reads an explicit `[]`
as a deliberate "no capabilities". Without it the `platform_ops` template enrolls
with zero capabilities ticked, so it is bundled here rather than left as a
prerequisite PR — the zero-click default is the template's headline behaviour. The
field is exposed unfiltered (it states what the template *declares*), and the
frontend narrows it through the already-`can_use`-filtered advertised set, so a
default the team is not enabled for is neither pre-ticked nor rendered.

**Template-declared reasoning defaults (#2473), so the ops agent actually reasons.**
`platform_ops` is a diagnostic agent whose turns are inherently multi-step, but
REASON-01 level 3 and Amendment B were settable per *instance* only — every operator
would have had to find the switch in the form. Amendment B's own title says "let an
agent AUTHOR preselect reasoning ON" (#2175); the author had no surface to do it
from, and the SDK's `AgentTuning` did not even declare `reasoning_default_on`.

Closed by mirroring the `default_capability_ids` path end to end: `AgentDefinition`
declares both fields, `_definition_to_agent_tuning` projects them onto the
`default_tuning` already on the wire, `AgentTemplateSummary` exposes them, and the
form seeds from the template instead of a hardcoded `false` pair. No new transport
and no migration — `ManagedAgentTuning` already carried both with `False` defaults,
so stored rows and older pods deserialize unchanged.

**A seed, not a gate.** The pre-tick is unconditional, which copies existing
behaviour rather than adding any: the form's Reasoning card already renders
unconditionally and never reads `reasoning_enabled_model_ids`. Levels 1-2 stay
enforced live on the send path, so a declared `True` is inert where no model has
reasoning enabled — exactly as a hand-ticked `True` is today, and the control
reappears if an admin enables one later. Enrollment still overwrites both from the
request, so the operator's choice wins once the form has been seen.

**A 30-call per-turn budget for this agent only.** `platform_ops` gets its own
`ToolSelectionPolicy`, leaving the shared `REASONING_SAFE_TOOL_SELECTION` (12) to
the five conversational agents. An ops agent works unattended — the operator asks a
question and walks away — so one turn is the whole investigation, not one exchange,
and as this family grows past Postgres a single investigation crosses several
services. The cap's `exit_behavior="continue"` means running out does not fail the
turn: the agent silently answers from partial evidence, which on a diagnostic agent
is the dangerous failure, since a confident answer built on half the data is what an
operator will act on.

This also retires a stale justification: the #1780 duplicate-tool-call defect that
once made reasoning risky here was root-caused and fixed — stripping the model's own
reasoning between loop steps left it re-deriving the same plan, so
`CheckpointHygieneMiddleware` threads the open turn's reasoning back as text. The
tool-call cap never addressed it. `support/thinking.py` still described that fix as
"candidate ... not yet wired" long after it shipped; corrected here.

---

## 4. Proof of quality

**Tests added or updated:** 44 new tests across 5 files for the capability, plus WP1's control-plane and `AgentFormModal` tests.

| `test_multi_statement_string_is_rejected`, `test_single_statement_writes_are_rejected`, `test_read_only_escape_script_is_rejected`, `test_single_set_statement_cannot_unlock_later_writes` | `libs/fred-runtime/tests/test_platform_sql_postgres_integration.py` | The executable form of the three security claims above, against a **real Postgres** (gated behind a DSN env var) |
| `test_row_cap_truncates_and_flags`, `test_exactly_200_rows_is_not_flagged`, `test_statement_timeout_maps_to_timed_out`, `test_with_cte_works`, `test_explain_without_analyze_works`, `test_undefined_column_error_carries_server_message`, +2 | same file | Row cap boundary, timeout, that legitimate CTE/EXPLAIN still work, server message preserved |
| `test_clamp_timeout_s`, `test_postgres_config_builds_dedicated_asyncpg_engine`, `test_sqlite_escape_hatch_builds_no_adapter`, `test_missing_password_fails_loud`, `test_non_server_error_never_leaks_topology`, `test_pool_checkout_timeout_maps_to_a_retry_unchanged_message`, +5 | `libs/fred-runtime/tests/test_platform_sql.py` | Adapter contract offline: clamping, dedicated pool, fail-loud config, **no DSN/host/driver repr leaking into error messages** |
| `test_giant_cell_truncated_in_place_with_marker`, `test_total_cap_drops_trailing_whole_rows_never_mid_cell`, `test_run_query_empty_result`, `test_server_error_returns_is_error_result_with_server_message`, `test_missing_port_fails_loud`, `test_tool_signatures_carry_only_llm_arguments`, +10 | `libs/fred-capability-platform-ops/tests/test_platform_postgres.py` | Both tools, every truncation edge, `is_error` degradation, and that tool signatures carry **only** LLM arguments |
| `test_capability_registers_via_entry_point_and_boots`, `test_manifest_declares_the_minimal_tool_surface` | `libs/fred-capability-platform-ops/tests/test_registry.py` | Entry-point registration + boot invariant |
| `test_platform_ops_is_registered`, `test_platform_ops_prompt_loads_non_empty`, `test_platform_ops_defaults_to_platform_postgres` | `apps/fred-agents/tests/test_platform_ops_template.py` | Template registration and its default capability list |

```
All checks passed (fred-core, fred-sdk, fred-runtime, writable-document verified
locally; full sweep + all remaining modules verified green by the push CI on swift
for this exact tree — "Code Quality: Push on swift" ✅)
```

**Raw `basedpyright` output:**

```
libs/fred-sdk:     0 errors, 0 warnings, 0 notes
libs/fred-runtime: 0 errors, 1 warning, 0 notes
  fred_runtime/__main__.py:80:5 - warning: Code is unreachable (reportUnreachable)
  ^ pre-existing on swift, untouched by this branch
libs/fred-capability-platform-ops: baseline files added (.baseline/*.json)
```

**`make test` output:**

```
Verified by the push CI on swift for this exact tree — "Backend Tests" green
across fred-sdk, fred-runtime, fred-agents and the new
libs/fred-capability-platform-ops matrix entry.
```

---

## 5. Docs updated

| ------------ | ------------ |
| Backlog `[ ]` item is now done | n/a — `BACKLOG.md` frozen; #2458 is the tracking unit |
| New behaviour, API field, or contract change | `docs/swift/capabilities/AUTHORING.md` — capability row for `PlatformPostgresCapability` |
| Frozen contract touched | `docs/swift/design/CONTROL-PLANE-PRODUCT-CONTRACT.md` — `default_capability_ids` on `AgentTemplateSummary` (WP1). Otherwise additive: the new optional `RuntimeServices.platform_sql` field is appended last for positional safety; no existing field or signature changed |
| UX component implemented or visual status changed | n/a — no new UI; en/fr i18n keys only (`capability.platform_postgres.*`) |
| WORKPLAN sprint item finished | n/a — `WORKPLAN.md` frozen |
| Code and a design doc now diverge | No — `PLATFORM-POSTGRES.md` `Status:` updated to `implemented 2026-08-27 (#2458)`, with the two genuinely-open items (§4 result-format A/B, §7.3 agent-level evaluation) kept as explicit follow-ups |
| Frozen contract touched (#2473) | `docs/swift/design/CONTROL-PLANE-PRODUCT-CONTRACT.md` §33 addendum (2026-08-28) — the two reasoning fields on `AgentTemplateSummary` and the seed-not-gate semantics; `docs/swift/design/RUNTIME-EXECUTION-CONTRACT.md` §8.59 — the SDK/`AgentTuning` fields and the `_definition_to_agent_tuning` projection |
| Recorded design and code diverged (#2473) | Fixed rather than documented around: Amendment B always said "author", so the code now matches the RFC instead of the doc being trimmed to match the code |

---

## 6. Close-out statement

```
## Task close-out
- Code: new PlatformSqlPort in fred-sdk; fred-runtime adapter with dedicated pool
  and three-layer server-side read-only enforcement; libs/fred-capability-platform-ops
  with the platform_postgres capability (2 tools, 1 config field, truncating JSON
  envelope); platform_ops ReAct template + prompt + en/fr i18n; root Makefile and CI
  matrix wiring.
- Tests: 44 added across 5 files — capability unit tests with a stubbed port, adapter
  contract tests offline, and a real-Postgres integration suite proving the three
  read-only rejections. Green in CI.
- Docs updated: docs/swift/rfc/ADMIN-OPS-AGENTS-RFC.md (new),
  docs/swift/rfc/admin-ops-capabilities/PLATFORM-POSTGRES.md (new),
  docs/swift/capabilities/AUTHORING.md (capability row).
- Also in this PR: #2473 — template-declarable reasoning defaults across fred-sdk,
  fred-runtime, control-plane and the agent form (7 tests added); platform_ops
  declares both plus its own 30-call per-turn tool budget.
- Tracking: #2458 and #2473 (milestone swift-v2.2) — both close on merge.
- Skipped steps: none. Steps 1-3 predate implementation (spec signed off 2026-08-27).
  Note the process deviation this PR corrects: the work initially reached swift
  without a PR and has been reverted there (517b05fd) to re-land through review.
```

---

## 7. Risk and rollback

Blast radius is small by construction. The capability is `ADMIN_GATED` and invisible
to every team until an operator grants it — the admin team roster is the entire trust
boundary, and there is deliberately no table filtering, because operators grant this
only to teams whose members already hold direct `psql` access. Nothing existing
changes behaviour: `RuntimeServices.platform_sql` is a new optional field, and a
missing port makes the tools fail loud rather than degrade silently.

The two realistic failure modes:

1. **A read-only escape** — the one that would actually matter. Mitigated by three
   independent server-side layers and proven by real-Postgres tests, including the
   specific `SET default_transaction_read_only = off; COMMIT; DELETE` escape attempt.
2. **Pool exhaustion** — bounded to this feature: the adapter owns a separate pool of
   2 connections with no overflow, so the worst case is that platform-ops queries
   queue among themselves. The app engine's pool, and therefore the checkpointer and
   history hot path, is untouched.

Deliberately no HITL: read-only plus server-enforced means approval friction buys no
security while defeating the "chat instead of shelling in" purpose.

**How do we roll back?**

Revert this PR's merge commit — the change is additive, owns no tables, and ships no
migration, so a revert is complete. Without a code change, an operator can also just
revoke the capability grant, which removes it from every agent immediately.


florian-muller:
Related: #2473 — let an agent template declare its reasoning defaults (REASON-01 level 3 + Amendment B).

The `platform_ops` template added in this PR should ship offering reasoning, and starting in Boost, without every operator rediscovering that in the agent form. Today neither is possible: `reasoning_enabled` / `reasoning_default_on` are settable per *instance* only — `_definition_to_agent_tuning` doesn't project them, and the frontend hardcodes both to `false` on template select.

#2473 mirrors the `default_capability_ids` path end to end so the template can declare both, then sets them on `platform_ops`.

dimitri-tombroff:
Florian do you want me (us) to validate and merge this ? Looks like as soon merged the better no ?

florian-muller:
> Florian do you want me (us) to validate and merge this ? Looks like as soon merged the better no ?

Yes, it's a first version, ready to merge ! Only with postgres capability. I am working on the opensearch part on an other branch (for an other PR). 

I will resovle the conflcit and merge it so its in the next PR :)

PR #2508, opened 2026-09-03
feat(#2506): delete a team from the admin registry behind a critical-action dialog

### LINKED ISSUE #2506: Admin > Teams: delete a team from the registry, behind an irreversible-action confirmation
## Problem

`DELETE /control-plane/v1/teams/{team_id}` already exists and is `platform_admin`-gated
(`teams/service.py::delete_team` - drops every ReBAC relation referencing the team, then
the `team_metadata_store` row). It has no UI. The admin Teams page is create-only: its
header comment still states "this page has no edit/delete affordance for a team, only
creation and registry visibility".

So the only way to remove a stale team today is a hand-rolled API call, which nobody does -
dead teams accumulate in the registry and pollute the marketplace and the KPI rankings.

## Proposal - frontend only

Add a per-row delete action to the existing-teams table in `AdminTeamsPage`, gated behind
the app's standard critical-action confirmation.

- Reuse `useConfirmationDialog` with `criticalAction: true` - the same formalism
  `TeamAgentsPage` uses for instance deletion (filled "Cancel", text-only error-coloured
  "Delete"). No new dialog component.
- The dialog names the team and states plainly that the action is irreversible: the team's
  registry entry and every access relation attached to it are dropped and cannot be
  restored.
- Wire `deleteTeamControlPlaneV1TeamsTeamIdDelete` (already in the generated client) with a
  friendly alias in `controlPlaneApiEnhancements.ts`, invalidating
  `ControlPlaneTeam/{id}` + `ControlPlaneTeam/LIST` like the sibling team mutations do.
- Success/error toasts through `useMutationAction` + `useApiErrorToast`, matching the
  create-team flow already on the page (403 -> "not allowed", 404 -> already gone).

## Out of scope

- No backend change. The endpoint, its authorization and its relation cleanup are done.
- No cascade of the team's *data* (documents, sessions, agent instances). The endpoint
  deletes the registry entry and the relations; broadening that is a separate discussion.

## Acceptance

- A platform admin sees a delete action on each row of "Existing teams".
- Confirming removes the team and the table refreshes without a manual reload.
- Dismissing the dialog changes nothing.
- A non-admin's attempt surfaces the forbidden toast rather than a silent failure.
- Unit test covers: dialog shown before any call, mutation fired on confirm, no call on
  cancel.

simcariou:
Closes #2506.

## What

The admin Teams table gains a per-row delete action. `DELETE /teams/{team_id}`
has existed and been `platform_admin`-gated since AUTHZ-05 with no UI, so the
only way to drop a stale team was a hand-rolled API call.

**Frontend only** - no backend change. The endpoint, its authorization and its
relation cleanup were already in place.

## How

- `AdminTeamsPage` gets a third, fixed-width column holding a `delete` icon
  button (fixed track, not `auto`: `DataTable` resolves header and body tracks
  as separate grids, and this header is empty).
- The click opens the shared `useConfirmationDialog` with `criticalAction: true`
  - the same inverted-emphasis formalism leave-team and delete-agent already
  use (filled/primary Cancel, text/error Confirm). No new dialog component.
- The message names the team and states that the registry entry and every
  access relation attached to it go with it, with no undo.
- `deleteTeamControlPlaneV1TeamsTeamIdDelete` gets a friendly alias in
  `controlPlaneApiEnhancements.ts` and the same `ControlPlaneTeam/{id}` +
  `ControlPlaneTeam/LIST` invalidation as its sibling team mutations, so the
  table refreshes on its own.

## Tests

New `AdminTeamsPage.test.tsx`, 5 cases, mounting the **real**
`ConfirmationDialogProvider` rather than stubbing it - the guarantee under test
is that the request cannot leave the client unconfirmed, so the dialog users
actually see is what gets driven:

- a bare click fires no request and opens the dialog;
- the dialog names the team and carries the irreversibility copy;
- confirming deletes only that team and toasts success;
- cancelling fires nothing;
- a failed deletion surfaces an error toast, not silence.

## Verification

- `vitest` — 5/5 pass
- `tsc --noEmit` — clean
- `eslint` + `prettier` on the touched files — clean

Root `make code-quality` not run here (cold worktree); the CI matrix covers it.

## Notes for review

- `list_all_teams` already filters personal spaces out of the registry view, so
  a personal space is not reachable from this table.
- Scope stops at the registry: the endpoint drops the row and the relations, not
  the team's documents/sessions/agent instances. Broadening that is a separate
  discussion, not this PR.

### DIFF
--- a/apps/frontend/src/locales/en/translation.json
+++ b/apps/frontend/src/locales/en/translation.json
@@ -2261,13 +2261,14 @@
         "empty": "No teams yet.",
         "table": {
           "name": "Team",
-          "admins": "Admin(s)"
+          "admins": "Admin(s)",
+          "actions": "Actions"
         }
       },
       "deleteTeam": {
         "action": "Delete team {{name}}",
         "dialogTitle": "Delete {{name}}?",
-        "dialogMessage": "This permanently removes the team from the registry, along with every access relation attached to it. Members lose access immediately. This cannot be undone.",
+        "dialogMessage": "{{name}} is removed from the registry and every access relation involving it is dropped. Members lose access immediately, and the team's document libraries, agents and prompts are left with no owner - their content stays on disk but nobody can reach it. This cannot be undone.",
         "confirm": "Delete team",
         "cancel": "Cancel",
         "successSummary": "Team deleted",
--- a/apps/frontend/src/locales/en/translation.json
+++ b/apps/frontend/src/locales/en/translation.json
@@ -2264,6 +2264,19 @@
           "admins": "Admin(s)"
         }
       },
+      "deleteTeam": {
+        "action": "Delete team {{name}}",
+        "dialogTitle": "Delete {{name}}?",
+        "dialogMessage": "This permanently removes the team from the registry, along with every access relation attached to it. Members lose access immediately. This cannot be undone.",
+        "confirm": "Delete team",
+        "cancel": "Cancel",
+        "successSummary": "Team deleted",
+        "errors": {
+          "summary": "Failed to delete team",
+          "fallbackDetail": "Could not delete the team. Please try again.",
+          "forbiddenDetail": "You are not allowed to delete teams."
+        }
+      },
       "createTeam": {
         "title": "Create team",
         "nameLabel": "Team name",
--- a/apps/frontend/src/locales/fr/translation.json
+++ b/apps/frontend/src/locales/fr/translation.json
@@ -2261,13 +2261,14 @@
         "empty": "Aucune équipe pour l'instant.",
         "table": {
           "name": "Équipe",
-          "admins": "Administrateur(s)"
+          "admins": "Administrateur(s)",
+          "actions": "Actions"
         }
       },
       "deleteTeam": {
         "action": "Supprimer l'équipe {{name}}",
         "dialogTitle": "Supprimer {{name}} ?",
-        "dialogMessage": "L'équipe est définitivement retirée du registre, ainsi que toutes les relations d'accès qui la référencent. Les membres perdent l'accès immédiatement. Cette action est irréversible.",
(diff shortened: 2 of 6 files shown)


simcariou:
## Review round applied (011e91d9b)

Ran `/code-review` on the diff before calling this done. Nine findings, all applied. The one that mattered:

**The dialog copy understated what deletion does.** `delete_all_relations_of_reference` filters on `tup.key.user == team:X **or** tup.key.object == team:X`, and a document library is owned via `tag:<id>#owner -> team:<id>` — the team is the *subject* there. So deleting a team also drops the ownership of its libraries, agents and prompts: the Postgres rows, vectors and blobs survive with no owner and no way back. "Removes the team from the registry, along with every access relation attached to it" read as an access-control cleanup, which is exactly the understatement that makes a destructive action feel safe. The copy now states it.

That gap is pre-existing and backend-side, so it is **not** fixed here — it is filed as #2511, which lays out three options (refuse to delete a non-empty team / cascade / reassign) and recommends the first. This PR's dialog is mitigation, not the fix.

Also applied:
- cache invalidation widened to the per-team member/session/agent/prompt lists — the store refreshes only through tags, so those kept serving a dead team's rows;
- the row action is disabled while a delete is in flight (the dialog closes before the request settles, so a second confirm 404s and toasts a failure for a deletion that succeeded);
- `rowKey` passed, per `DataTable`'s own prop contract now that rows carry a destructive action;
- the action column has a real header label instead of `""`, which is both DataTable's sort identity key and the header's accessible name;
- dropped a `useCallback` that implied a stability the unmemoized `ConfirmationDialogProvider` context cannot provide;
- the test mock records the deletion on the **trigger call**, not inside `unwrap()`. That was the real one: as written, the suite's headline guarantee was an assertion about awaiting, not about dispatching.

Verified by mutation: bypassing the dialog fails 4 of 5 tests, dropping `.unwrap()` fails 1. Before the fix, a dialog bypass would have passed.

Still green: vitest 5/5, `tsc --noEmit`, eslint, prettier.

PR #2509, opened 2026-09-03
feat(#2507): label the top-agents KPI series with the owning team

### LINKED ISSUE #2507: Analytics: label the 'most talked-to agents' KPI series with the owning team
## Problem

The "Most talked-to agents" chart on the admin Analytics tab labels each series with the
agent instance's display name alone. Agent instances are per-team, and their names are not
unique across the platform - several teams routinely run an instance called "Assistant",
"Analyst" or "Support".

`top_agents_by_conversations.py` already knows this: it counts colliding names and
disambiguates them with a truncated instance id suffix, e.g. `Analyst (3f9c1ab2)`. That
keeps the lines distinguishable but tells the reader nothing useful - a platform admin
reading a cross-team ranking wants to know *which team* the agent belongs to, and an opaque
8-char id is strictly worse than the answer we already have in the data.

## Proposal - backend preset + i18n

Label each series with its owning team, and drop the id-suffix fallback.

- `agent.turn_completed` already carries `dims.team_id` (the preset is `team_scopable=True`
  on that basis). The phase-1 `latest_name` top_hits sub-agg already reads
  `dims.agent_instance_name` from the most recent event - add `dims.team_id` to the same
  `_source.includes`, so this costs no extra query and keeps the deleted-instance safety
  net.
- Resolve ids to display names with `_resolve_team_names`, which
  `top_teams_by_sessions.py` already implements against `team_metadata_store` (with a
  fall-back to the raw id on any error). **Move it to a shared helper and have both presets
  import it** rather than copying it - per the prime directive, this converges the two call
  sites instead of adding a third variant.
- Label becomes `"<agent name> - <team name>"`. Team names are unique in the registry, so
  the id suffix becomes unnecessary: two same-named instances in different teams are now
  distinguished by the team. Keep a short id suffix only for the residual case of two
  same-named instances *inside the same team*.
- The instance id stays the accumulation key throughout - the label is display-only. That
  invariant is what stops two instances merging into one line, and this change must not
  weaken it.

## Out of scope

- No new preset, no new endpoint, no frontend chart change: `MultiSeriesLineChart` renders
  whatever `series` the backend returns. Only the English/French chart subtitle wording
  needs a look.
- The per-team-scoped view of this same preset (when `team_id` is passed) shows one team by
  definition - the team suffix is redundant there and should be omitted.

## Acceptance

- The cross-team chart shows `Analyst - swiftpost` / `Analyst - fredlab` rather than
  `Analyst (3f9c1ab2)` / `Analyst (7b02de41)`.
- A team-scoped request keeps bare agent names.
- A team whose registry entry has since been deleted falls back to its id, not a crash.
- `test_kpi_top_agents_by_conversations.py` extended for the new labelling, including the
  same-name/same-team collision case.

simcariou:
Closes #2507.

## What

The "Most talked-to agents" chart on the admin Analytics tab labelled each
series with the agent instance's display name alone. Instance names are
per-team and routinely collide - several teams run an "Assistant", an "Analyst"
or a "Support". The preset already knew this and disambiguated collisions with
a truncated instance id (`Analyst (3f9c1ab2)`), which keeps the lines apart but
answers the wrong question: in a cross-team ranking a platform admin wants the
owning team, and it was already in the data.

## How

- Phase 1's existing `top_hits` sub-agg (renamed `latest_name` -> `latest_dims`,
  since it now reads more than the name) adds `dims.team_id` to its
  `_source.includes`. No extra query, and the deleted-instance safety net is
  unchanged: both values were persisted at emit time, so a since-deleted
  instance still labels its own line.
- Labels become `"<agent name> - <team name>"`. The id suffix survives only as
  the residual escape hatch - two same-named instances *inside one team*, or an
  event that carried no `team_id`.
- A **team-scoped** request keeps bare names: every series would otherwise carry
  the same suffix, repeating the filter the caller already set.
- The instance id stays the accumulation key throughout; the label is
  display-only. That invariant is what stops two instances collapsing into one
  summed line, and this change does not touch it.

## Reuse, not duplication

`top_teams_by_sessions` already had a `_resolve_team_names` helper against
`TeamMetadataStore` with an id fallback. Rather than copy it into a second
preset, it moved to `kpi.utils.resolve_team_names` and both presets now import
it. Net: one implementation where there would have been two.

## Tests

`test_kpi_top_agents_by_conversations.py` grows from 5 to 9 cases. The
team-registry lookup is faked with an autouse fixture, in the same spirit as the
existing fake OpenSearch client:

- two same-named instances in different teams -> distinguished by team;
- two same-named instances in the *same* team -> id suffix still applied;
- unique names are still team-qualified (the suffix is the normal display);
- a team-scoped request keeps bare names;
- an event with no `team_id` falls back to the bare name;
- a deleted team falls back to its raw id rather than dropping the line.

The pre-existing pinning/accumulation/empty-result cases are unchanged.

## Verification

- `pytest -k kpi` — 112 pass
- `ruff format` + `ruff check` — clean

8 failures in `test_main.py` / `test_team_visibility.py` /
`test_get_team_by_id_projection_budget.py` were confirmed **pre-existing** by
re-running them against a stashed tree on the same base. Root `make
code-quality` not run here (cold worktree); the CI matrix covers it.

## Notes for review

- No API contract change: same route, same `MultiSeriesTimeSeriesResponse`. Only
  the string values inside `series`/`rows` change, so no client regeneration.
- No frontend change either - `MultiSeriesLineChart` renders whatever `series`
  the backend returns, and its legend is already a `maxHeight: 44` scroll area.
  Worth a glance at review time, since labels do get longer.
- Contract note added to `CONTROL-PLANE-PRODUCT-CONTRACT.md` §36.

### DIFF
--- a/apps/control-plane-backend/control_plane_backend/kpi/presets/team_names.py
+++ b/apps/control-plane-backend/control_plane_backend/kpi/presets/team_names.py
@@ -1,0 +1,55 @@
+# Copyright Thales 2026
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Team-id to display-name resolution, shared by the presets that label a team.
+
+Its own module rather than `kpi/utils.py`: that one is a stdlib-only leaf every
+preset imports for `resolve_interval`, and the container this needs drags in
+keycloak and sqlalchemy.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from fastapi import Request
+from fred_core.common import TeamId
+
+from control_plane_backend.app.dependencies import get_application_container
+
+logger = logging.getLogger(__name__)
+
+
+async def resolve_team_names(request: Request, team_ids: list[str]) -> dict[str, str]:
+    """Return {team_id: display_name}, falling back to the id on any error.
+
+    A team's name lives in `team_metadata_store`, not in Keycloak (AUTHZ-05
+    review item 9).
+    """
+    if not team_ids:
+        return {}
+    try:
+        store = get_application_container(request).get_team_metadata_store()
+        metadata_by_id = await store.get_by_team_ids([TeamId(tid) for tid in team_ids])
+        return {
+            tid: metadata_by_id[TeamId(tid)].name
+            if TeamId(tid) in metadata_by_id
+            else tid
+            for tid in team_ids
+        }
+    except Exception:
+        # Logged, not silent: an unreachable store and a genuinely deleted team
+        # both render as raw ids, and only this line tells the two apart.
+        logger.warning("Falling back to raw team ids for KPI labels", exc_info=True)
+        return {tid: tid for tid in team_ids}
--- a/apps/control-plane-backend/control_plane_backend/kpi/presets/top_agents_by_conversations.py
+++ b/apps/control-plane-backend/control_plane_backend/kpi/presets/top_agents_by_conversations.py
@@ -21,7 +21,7 @@ from typing import Any
 
 from fastapi import Request
 from fred_core import KeycloakUser
-from fred_core.common import TeamId
+from fred_core.common import TeamId, is_personal_team_id
 from fred_core.kpi.opensearch_kpi_store import OpenSearchKPIStore
 
 from control_plane_backend.kpi.presets.base import PresetDef
@@ -29,7 +29,8 @@ from control_plane_backend.kpi.presets.common import (
     MultiSeriesPoint,
     MultiSeriesTimeSeriesResponse,
 )
-from control_plane_backend.kpi.utils import resolve_interval, resolve_team_names
+from control_plane_backend.kpi.presets.team_names import resolve_team_names
+from control_plane_backend.kpi.utils import resolve_interval
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


simcariou:
## Review round applied (27c3d747a)

Ran `/code-review` on the diff before calling this done. Two findings were real bugs, and one correction to my own PR description.

**Personal spaces leaked a user id into the chart.** `dims.team_id` on `agent.turn_completed` is the turn's *scope*, which is `personal-<user_uid>` for personal-space chats. A personal team is synthetic — `build_personal_team` constructs it in memory, there is no `TeamMetadataRow` — so the name lookup fell through to the raw id and a platform admin would have seen `"MyAgent - personal-3f9c1ab2-4d5e-…"` in a legend. Personal scopes are now never qualified, covered by a test.

**`kpi/utils.py` was the wrong home for the shared helper.** It is a stdlib-only leaf that ten presets import for `resolve_interval`; the container the helper needs pulls keycloak and sqlalchemy. Measured against `origin/swift` with the same venv: 74 modules / 0.014s → 2619 modules / 2.1s, which also made `test_kpi_trend_utils.py` (pure date math) depend on the Keycloak client being importable. The helper moved to its own `kpi/presets/team_names.py`; `utils.py` is now byte-identical to `swift` and measures back at 74 modules.

I got this wrong twice before landing it: my first "baseline" measurement stashed only my *uncommitted* edits, so it compared my heavy version against my heavy version and appeared to exonerate the change. The real baseline came from `git archive origin/swift`.

Also applied:
- `resolve_team_names` swallowed every exception with no log line, so a metadata-store outage and a genuinely deleted team rendered byte-identically. It now warns, and has 4 tests — it had none anywhere in the repo, in either caller, before this PR doubled its callers.
- `_build_labels`'s `scoped` flag was dead code (an empty `team_names` already yields bare names) and forced the caller to write the same `team_id is not None` condition twice. Removed.
- `top_teams_by_sessions` kept a `logger` with no user after the helper moved out.
- a test for two colliding names with *no* `team_id` — the shape of every event emitted before the dim existed, and the case where the id-suffix mechanism actually earns its keep.

**Legend readability, which my PR description flagged and then shipped nothing for.** That was the wrong call: at `TOP_N = 10` with team-qualified labels, a 44px legend hides most of a chart whose entire purpose is a cross-team ranking. `MultiSeriesLineChart`'s cap goes to 88px. It is a max, not a height, so the two other consumers (1-3 series) render identically.

Green: 118 KPI tests, `ruff format`/`check`, and the 20 frontend tests across the three chart consumers.

Issue #2515, opened 2026-09-03
Chat composer: pasting a file from the clipboard does not attach it to the conversation


simcariou:
## Problem

The chat composer accepts attachments from two entry points only: the attach-file
picker (`attach_files` composer control) and drag & drop onto the page. Copying a
file in the OS file manager - or taking a screenshot - and pasting it into the
composer does nothing: the clipboard files are dropped and, at best, a file path
is pasted as text.

Every other entry point of the same feature already works, so the gap is only in
the composer's paste handling.

## Expected

Pasting into the chat composer with files in the clipboard attaches them to the
conversation exactly like the picker and drop paths do: same upload-warning gate,
same fast-ingest + session-attachment persistence, same chips, same send gating
while the ingestion runs.

## Notes / scope

- `useChatAttachments.addFiles(files, source)` already does all the work; the
  paste path only has to feed it. `source` gains a `"paste"` member.
- Only attach when the agent exposes the `attach_files` control, like drop does.
- A clipboard that carries real text next to a file must still paste the text:
  Word/Excel put a rendered PNG in the clipboard alongside the copied cells, and
  turning that into a silent image upload would be worse than the current gap.
  File-manager copies that leave a path/URI in `text/plain` still count as files.
- Pasted screenshots arrive as `image.png` for every paste. Since attachments are
  resolved by name by the document tools, give clipboard files a unique name.

Out of scope: paste in any other input, and multi-file clipboard ordering.

simcariou:
Browser limit found while testing the draft (#2517): Firefox exposes only the first file of a multi-file copy in `clipboardData.files` - [Mozilla bug 864052](https://bugzilla.mozilla.org/show_bug.cgi?id=864052), open since 2013, parity gap with Chrome and Safari which expose every file. A three-file copy logs `types=Files files=1 attached=1` on Firefox. Nothing page-side can recover the others; drag & drop and the picker remain the multi-file paths there. Recorded in `CHAT-COMPONENT-SPECS.md` §8.3.

PR #2523, opened 2026-09-03
fix(#2522): keep the attachment delete button out of the row button

### LINKED ISSUE #2522: fix(chat): attachment drawer nests the delete button inside the row button
## Problem

Each row of the session attachments drawer (\`SessionAttachmentsDrawer\`) is a \`<button>\` that opens the markdown preview, and it wraps the delete \`IconButton\`, which is itself a \`<button>\`. React flags this at render time:

```
In HTML, <button> cannot be a descendant of <button>.
This will cause a hydration error.
```

Invalid HTML aside, the delete handler has to call \`stopPropagation\` to keep the row from opening the preview it sits in.

## Fix

Restructure the row the way \`DocRow\` already does: the row becomes a \`<div>\`, the preview target becomes its own \`<button>\` filling the row up to the action buttons, and the delete \`IconButton\` sits beside it as a sibling. No \`stopPropagation\` needed anymore.

Regression test: the drawer renders no \`button button\`, clicking the row opens the preview, clicking delete calls \`onDelete\` without opening it.

Pre-existing (the row shape dates back to the chat page M3 polish, #2214); surfaced while validating #2515 in the browser.

simcariou:
Closes #2522

## What

Each row of the session attachments drawer was a `<button>` (opens the markdown preview) wrapping the delete `IconButton`, itself a `<button>`. React flags the nested button as invalid HTML and a hydration error.

## How

Same shape as `DocRow`: the row becomes a `<div>`, the preview target becomes its own `<button>` that fills the row up to the action buttons, and the delete control sits beside it as a sibling. The delete handler no longer needs `stopPropagation`.

Visually unchanged: the row keeps its border, background, hover state and layout; the whole line minus the delete button remains the preview target.

## Tests

`SessionAttachmentsDrawer.test.tsx` (happy-dom): no `button button` in the rendered drawer, clicking the row opens the preview, clicking delete calls `onDelete` without opening the preview. The first test is red on the previous markup.

Pre-existing bug, surfaced while validating #2515 in the browser. Independent of that branch.

### DIFF
--- a/apps/frontend/src/rework/components/shared/molecules/SessionAttachmentsDrawer/SessionAttachmentsDrawer.module.css
+++ b/apps/frontend/src/rework/components/shared/molecules/SessionAttachmentsDrawer/SessionAttachmentsDrawer.module.css
@@ -14,19 +14,33 @@
   display: flex;
   align-items: flex-start;
   gap: var(--spacing-xs);
-  cursor: pointer;
   border: 1px solid var(--outline-retreat);
   border-radius: var(--radius-s);
   background: var(--surface-container-low);
   padding: var(--spacing-xs);
-  width: 100%;
   color: var(--on-surface);
-  text-align: left;
 }
 
 .row:hover {
   border-color: var(--outline-muted);
   background: var(--surface-container-high);
+}
+
+/* Opens the preview: fills the row up to the action buttons so the whole line
+ * stays clickable, minus the controls that must not live inside a button. */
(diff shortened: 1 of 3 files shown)


simcariou:
Duplicate: the same change (574a4c6fb, byte-identical) already landed on feat/2515-paste-attachments in d3596452a, so #2517 carries the fix and closes #2522.

Issue #2531, opened 2026-09-03
Bound native Deep sub-agent concurrency for shared deployments


florian-muller:
## Parent

#2738 — native Deep batch processing.

## Native Deep scope superseding the custom POC (2026-09-18)

Current parent: #2738, Tier 1. The custom `run_subagent` path and recursive-depth estimates above are historical evidence, not the implementation target. Reuse this issue for bounded concurrency with Deep Agents' native `task` children rather than creating a second fan-out issue.

Measure the actual native graph, provider and shared HTTP pool behavior. Choose and document the concurrency/admission scope using the existing runtime/provider controls where possible. Preserve cancellation and release capacity after errors; demonstrate that one large batch task does not starve unrelated users. The current unbounded-POC decision is not a shared-deployment acceptance criterion.

- [ ] Document the actual native-child concurrency and nested-delegation behavior for the installed Deep version.
- [ ] Choose the appropriate bound/admission policy and test concurrent users with throttling, cancellation and failure.
- [ ] Keep provider retries (#2535) and total tool-call budgets (#2744) distinct from concurrent-child limits.
- [ ] Record the decision and performance evidence before enabling large fan-out on a shared deployment.

**Porting approach:** policy and implementation work; no custom-capability semaphore/depth machinery is selected for reuse.

<details>
<summary>Original custom-delegation POC report (historical)</summary>

## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` §5.5 (reopened 2026-09-03 by the #2525 performance review). Tier 1, decision item.

## Problem

RFC §5.5 accepted unbounded fan-out on the premise that `max_tool_calls_per_turn` bounds it. Code check (2026-09-03) shows that premise is wrong on two counts:

1. The knob is not `None` on most agents: `fred_agents/tool_pacing.py` sets 12 on the deployment's tool-using ReAct agents and `platform_ops.py` sets 30.
2. It maps to `ToolCallLimitMiddleware(run_limit=...)` in `react/middleware/frame.py`, which counts **per graph run**. A child is its own graph run, so the counter resets at every depth level.

[...]

Issue #2529, opened 2026-09-03
Sub-agent: child sources and ui_parts reach the parent's turn + fan-out perf review


florian-muller:
## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` §6.5 (decided 2026-09-03: keep `ui_parts`). Tier 1.

## What to build

A child is not a reduced agent: a research child's citations and a document- or PowerPoint-producing child's parts reach the user on the parent's tool-result line. The tool returns `content_and_artifact` with a tool invocation result carrying the child's text plus its `sources` and `ui_parts`; the existing tool-result runtime event already transports both, so no frontend work is expected.

The invoker today fills only `content` on the invocation result; populating `sources` and `ui_parts` from the child's `final` event is new plumbing. Ownership and duplicate rendering of a child's parts inside the parent's turn are unresolved: record what the POC shows on this issue rather than pre-empting a design.

This ticket also closes the tier: N children are N full agent turns running concurrently in one pod, sharing model clients and connection pools, so it carries the tier's mandatory performance review of concurrent fan-out.

## Acceptance criteria

- [ ] A child that cites documents shows those citations on the parent's tool-result line.
- [ ] A child that produces a document renders it in the parent's turn; observations on ownership / duplicate rendering recorded here.
- [ ] `fred-performance-reviewer` run against concurrent fan-out (several children in one message) and findings addressed or recorded.
- [ ] `RUNTIME-EXECUTION-CONTRACT.md` §14 entry updated for `sources` / `ui_parts` on the invocation result.
- [ ] `make code-quality` + `make test` green; `/code-review` run.

## Blocked by

- #2525 — Sub-agent POC

florian-muller:
## Shipped (branch `feat/2529-subagent-ui-parts`, not pushed)

`LocalRegistryAgentInvoker` now fills `sources` and `ui_parts` on
`AgentInvocationResult` from the callee's `final` payload
(`libs/fred-runtime/fred_runtime/app/agent_app.py:803-814`), and `run_subagent`
passes both straight onto its `ToolInvocationResult`
(`libs/fred-capability-subagent/fred_capability_subagent/capability.py:196-210`).
Types already matched — `FinalRuntimeEvent` and `AgentInvocationResult` both
declare `tuple[VectorSearchHit, ...]` / `tuple[UiPart, ...]`
(`libs/fred-sdk/fred_sdk/contracts/runtime.py:318-322`,
[...]

Issue #2528, opened 2026-09-03
Sub-agent: token spend measurable per child (agent.subagent_turn_completed)


florian-muller:
## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` §7. Tier 1.

## What to build

Sub-agent token spend becomes visible per child. Today child turns emit no `agent.turn_completed` (only the streaming path emits it) and the invoker discards the `token_usage` carried on the child's `final` event, so sub-agent spend would be invisible.

- The invocation result gains `token_usage` (additive), populated by the invoker from the child's `final` event.
- The capability emits a separate metric, `agent.subagent_turn_completed`, through the KPI writer, with the child's tokens as quantities and the **parent's** session id, agent instance id and exchange id plus `invocation_depth` as dims. Separate, not folded into the parent's turn, so a runaway child is diagnosable.
- Ships with its Grafana panel and an `OBSERVABILITY-AND-AUDIT.md` note stating that queries reading `agent.turn_completed` alone under-count.

## Acceptance criteria

- [ ] `token_usage` is populated on the invocation result for a child run.
- [ ] `agent.subagent_turn_completed` is emitted once per child with the dims above.
- [ ] The metric is visible in Grafana on its own panel.
- [ ] `OBSERVABILITY-AND-AUDIT.md` documents the metric and the under-count caveat.
- [ ] `make code-quality` + `make test` green; `/code-review` and `fred-performance-reviewer` run (KPI emission is hot-path).

## Blocked by

- #2525 — Sub-agent POC

Issue #2527, opened 2026-09-03
Sub-agent: decide prompt mode (replace vs append) via system_prompt override


florian-muller:
## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` §5.2 and §6.7. Tier 1. This is the RFC's only remaining open question (§12).

## What to build

Decide how a child gets its instructions, by trying both modes on real agents:

- **Mode A (append):** framing first, then the parent's authored template kept verbatim as inherited instructions; the parent's `prompt` is the child's message. Already what the POC ticket ships.
- **Mode R (replace):** layer 1 of the composed system prompt becomes framing + the parent's `prompt`. Requires an optional `system_prompt` override on the invocation request, which replaces layer 1 only; layers 2–6 stay runtime-owned and cannot be removed by a caller.

The override field ships either way (mode R cannot be tried without it), and it is public: any Graph node calling `invoke_agent` can now run a registered callee with its template replaced, guardrails kept. The contract entry must say this in words.

Evaluation across at least three agents whose templates carry persona, output-language or business rules: retention of those rules in the child, how often the child asks questions or addresses a user, prompt-token overhead per child.

## Acceptance criteria

- [ ] `system_prompt` override exists on the invocation request and replaces layer 1 only.
- [ ] Both modes evaluated against ≥ 3 agents; results recorded on this issue.
- [ ] One mode kept; the other deleted from the RFC and the §5.2 layer table updated.
- [ ] `RUNTIME-EXECUTION-CONTRACT.md` §14 entry documents the override and the caller power it grants.
- [ ] `make code-quality` + `make test` green; `/code-review` run.

## Blocked by

- #2525 — Sub-agent POC

florian-muller:
Mechanism shipped on branch `feat/2527-subagent-prompt-mode` (to be merged into `add-subagents`): `system_prompt` override on the invocation request (replaces layer 1 only) and a `prompt_mode` capability config field (`append` default / `replace`) so both modes can be evaluated per agent.

[...]

PR #2536, opened 2026-09-03
feat(SUBAGENT-01): run_subagent — a ReAct agent delegates to fresh-context copies of itself

### LINKED ISSUE #2525: Sub-agent POC: run_subagent runs a fresh-context copy of the calling agent
## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` (§5, §6.1–6.4, §6.6). Tier 1 of the RFC's delivery tiers.

## What to build

A ReAct agent with the `subagent` capability enabled can delegate a prompt with one tool, `run_subagent`, and gets the child's answer back on an ordinary tool-call line. The child is the same agent (same `agent_id`, same capability selection and config as the parent), runs with a fresh context and no checkpointer, and never sees the parent's conversation.

Scope of this first slice, chosen so there is something to play with as early as possible:

- New standalone package `libs/fred-capability-subagent` (capability id `subagent`), one entry-point line, depending only on `fred-sdk`.
- The invoker privately carries the parent's resolved tuning, capability registry, team settings, agent instance id, exchange id and reasoning-enabled model ids, and forwards them only when the child `agent_id` equals the parent's. Today a child run through the invoker has no tools at all.
- The child runs without a checkpointer; it keeps the parent's `session_id` in its portable context for KPI grouping. Turn options are inherited (they narrow, never widen).
- Depth: `invocation_depth` on `CapabilityContext` (additive, default 0), the counter carried privately on the invoker (never on the request), the tool simply not returned by `tools()` at `max_depth` (clamped config, default 3, ceiling 5). No second limit in the invoker.
- Prompt mode A (append): a short sub-agent framing plus the parent's `prompt` travels as the child's message. No invocation-contract change in this slice; mode R is evaluated in a follow-up ticket.
- A failing child returns a tool error carrying the real message: the invoker maps `execution_error`, which it does not handle today.
- Child content is capped by a module constant; over the cap the child gets a "be more concise" error result, never a silent truncation.
- The tool description is built per turn and tells the parent to batch independent work into one message (parallelism is already real through LangGraph's tool node). It must never contain per-turn values (cache rule, RFC §6.3).

Not in this slice: HITL stripping (follow-up ticket; until it lands, only enable the capability on agents with no approval-gated tools), token accounting, child sources / ui_parts, prompt mode R.

## Acceptance criteria

- [ ] An agent with the capability enabled runs a sub-agent and receives its answer as text.
- [ ] The child has the parent's capabilities and config, and does not see the parent's conversation.
- [ ] No history row is written for a child; no phantom conversation appears in the session list.
- [ ] At `max_depth` the tool is absent from the child's schema; a forged request context cannot raise or reset the depth.
- [ ] A failing child returns a tool error carrying the real message, not an empty string.
- [ ] Several `run_subagent` calls in one assistant message run concurrently.
- [ ] Over-long child output yields the "be more concise" error result.
- [ ] `RUNTIME-EXECUTION-CONTRACT.md` §14 dated entry (depth, tuning propagation for same-agent children, `execution_error` fix); `capabilities/AUTHORING.md` documents `invocation_depth`.
- [ ] `make code-quality` + `make test` green; `/code-review` and `fred-performance-reviewer` run on the diff.

## Blocked by

- None (can start immediately).

florian-muller:
Tier 1 of #2520 — the POC that makes a ReAct agent able to delegate. Closes #2525.

Fred already had an agent-to-agent primitive (`AgentInvokerPort` / `LocalRegistryAgentInvoker`, `RUNTIME-EXECUTION-CONTRACT.md` §14), reachable only from Graph nodes. This adds the missing ReAct surface: one tool, `run_subagent`, that runs a fresh-context copy of the calling agent against a prompt the parent writes.

Design: [`docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md`](docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md). Shipped surface: `RUNTIME-EXECUTION-CONTRACT.md` §8.63.

## What's in it

**New package `libs/fred-capability-subagent`** — capability id `subagent`, one tool `run_subagent(prompt)`, one clamped config `max_depth` (default 3, ceiling 5). Installing the package *is* the registration (`fred.capabilities` entry point); it's wired into `apps/fred-agents` as an editable path dependency. `ADMIN_GATED`, and no agent selects it by default.

**A same-agent child inherits the parent turn.** Before this, a child invoked through `LocalRegistryAgentInvoker` reached the turn path with no `tuning`, no `capability_registry`, no `team_settings`, no `agent_instance_id` — so it had no capabilities and no tools at all. A child naming the same `agent_id` now inherits all of those, plus the parent's `turn_options` and the selections on its `RuntimeContext` (selected documents, search policy, language), so an agent narrowed to one folder gets children searching that folder. Cross-agent children are unchanged, byte for byte.

**And it runs the parent's own definition, not the registry template.** The registry holds templates; only `_apply_runtime_tuning` turns an instance's `selected_capability_ids` into the `default_mcp_servers` that reach `FredMcpToolProvider`. Forwarding `tuning` alone left the child with the native capability block but zero MCP tools, plus the wrong system prompt.

**All of it travels on a private attribute of the invoker** (`_ParentTurn`) — the `platform_chat_model_binding` doctrine (§8.55). Never on `RuntimeContext`, `PortableContext` or `AgentInvocationRequest`, so a caller can neither read nor forge it. The request's `PortableContext` becomes a *declaration* the invoker verifies: a same-agent child whose claimed `user_id`/`session_id`/`team_id` differs from the calling turn's is refused.

**Depth is bounded.** `CapabilityContext.invocation_depth` (additive), counted on the invoker, which re-enters at *d+1* on every invocation — cross-agent included, so an A → B → A cycle is bounded too. Enforcement is capability-side in `tools()`: at `max_depth` the tool is simply not returned, so a leaf child is never shown a delegation it would only be refused. One bound, one place — deliberately no second limit in the invoker.

**The child runs without a checkpointer.** `RuntimeServices` is per-request, so it's a per-run substitution, not a pod-level change. Without it the child would map the parent's `session_id` to LangGraph's `thread_id`, load the parent's checkpoint — seeing the whole conversation — and overwrite it mid-turn.

**Bug fix, affects every `invoke_agent` caller:** `LocalRegistryAgentInvoker` handled `final`, `assistant_delta` and `node_error` but not `execution_error` — the terminal event a raising child ends its stream with. Callers (the `fred-rags` pod included) got `is_error=True` with an **empty** message, which a model reads as "the callee answered nothing". It now returns the event's real message.

## Deliberately not in this PR

Each has its own issue and its own branch, to be merged into `add-subagents` before this goes green:

- **#2526** — HITL stripping. A child has no checkpoint an interrupt could persist to, so until this lands, enable `subagent` only on agents with no approval-gated tools.
- **#2528** — per-child token accounting (`agent.subagent_turn_completed`). Child turns emit no turn-level KPI today, so their spend is counted nowhere.
- **#2529** — child `sources` / `ui_parts` on `AgentInvocationResult`.
- **#2527** — prompt mode (replace vs append) via a `system_prompt` override. Needs a developer call on POC behaviour.
- **#2531** — the fan-out bound. **Depth bounds height; nothing bounds width.** N children are N full agent turns running concurrently in one pod against a shared connection pool. `max_tool_calls_per_turn` is not that bound (it's per graph run, and a child is its own graph run, so it resets at every level), and the per-child content cap doesn't compose either. Deliberate for the POC, to be settled with POC data.

MEMORY-02 / MEMORY-06 are **narrowed, not closed** — §14's known-gaps entries are updated to say exactly how far. #1859 is unaffected: `run_subagent` never sets `InvocationScope`.

## Verification

`make code-quality` and `make test` green. Hot path (N concurrent agent turns per pod) — `fred-performance-reviewer` and `/code-review` are still to run against the full branch once the follow-ups land.


florian-muller:
Closing this branch, finaly it bacame more a POC than a mergeable implementation. I am merging the good parts one by one as part of #2738

Issue #2618, opened 2026-09-10
Resources page fetches corpus type stats while the stats panel is collapsed


Maxime-DARAGON:
`TeamResourcesPage` fetches the corpus usage-by-type stats on every visit, even
though the cards that display them are collapsed by default.

The query is skipped on the active tab only:

```ts
const corpusStats = useGetCorpusTypeStatsKnowledgeFlowV1TagsStatsGetQuery(
  { teamId: fsTeamId },
  { skip: activeTab !== "resources" },
);
```

while the cards render behind `statsOpen`, whose initial value is `false`:

```ts
{statsOpen && activeTab !== "agents" && <ResourceStatsCards … />}
```

`GET /tags/stats` walks every library the user can read and every document in
each of them (see #2617), so this is a full corpus scan performed and discarded
on each page open. Measured at 124 ms on a 4-folder / 43-document team, and it
grows with both the folder count and the corpus size.

## Fix

Add `statsOpen` to the skip condition of the three stats queries (corpus, mine,
team). The existing `isUninitialized` guard in `onDocumentsChanged` already
covers the refetch path for a query that was never started.

Issue #2617, opened 2026-09-10
Resources page: GET /tags does one ReBAC lookup and one metadata scan per folder


Maxime-DARAGON:
Opening a team's Resources page is slow, and the cost grows with the number of
folders rather than with the number of documents shown.

## Measured

Opening the Resources page of a 4-folder / 43-document team on a local
prod-shaped stack (OpenFGA + Postgres) issues five knowledge-flow requests
totalling ~760 ms:

| request | time |
| --- | --- |
| `GET /tags?type=document&limit=1` (health probe) | 93 ms |
| `GET /tags?limit=10000&owner_filter=team&team_id=…` | 300 ms |
| `GET /tags/stats?team_id=…` | 124 ms |
| `GET /tasks?scope=team&kind=ingestion` | 231 ms |
| `POST /documents/metadata/tag-sizes` | 11 ms |

## Root cause

`TagService.list_all_tags_for_user` attaches `item_ids` to every returned tag
one tag at a time:

```python
for tag in sliced:
    item_service = get_specific_tag_item_service(tag.type)
    item_ids = await item_service.retrieve_items_ids_for_tag(user, tag.id)
```

For documents, `retrieve_items_ids_for_tag` goes through
`MetadataService.get_document_metadata_in_tag`, which per tag does:

- `rebac.lookup_user_resources(user, DocumentPermission.READ)` — an OpenFGA
  `ListObjects` over every document the user can read. The result does not
  depend on the tag, so N-1 of these calls are computed and thrown away.
  Measured at ~11 ms each on a 43-document corpus.
- `metadata_store.get_metadata_in_tag(tag_id)` — selects the full JSONB rows
  and hydrates their labels, only to read `document_uid` off each one.
  Measured at ~17 ms each.

`TagService.get_corpus_type_stats` repeats the same per-tag loop for its own
histogram.

Marginal cost is therefore ~28 ms per folder, on two endpoints, and both scale
with the total corpus size as well — a team with 25 folders pays roughly 1.4 s
of redundant work per page open.

## Proposed fix

Batch the fan-out, keeping the API contract unchanged:

- add `document_uids_by_tags(tag_ids)` and `metadata_in_tags(tag_ids)` to
  `BaseDocumentMetadataStore`, with a default per-tag loop for non-SQL stores
  and a single array-overlap query on `PostgresDocumentMetadataStore` — the
  same shape `total_size_by_tags` already uses;
[...]

PR #2622, opened 2026-09-10
perf(frontend): let the knowledge-flow cache survive a remount

### LINKED ISSUE #2619: knowledge-flow RTK Query slice has no cache: every mount refetches everything
The knowledge-flow RTK Query slice is configured so that nothing is ever served
from cache. The code already flags it:

```ts
// todo: remove this to improv app performance. Stale data should be handled with cache invalidation, not with timeouts.
keepUnusedDataFor: 0,
refetchOnMountOrArgChange: true,
refetchOnFocus: true,
refetchOnReconnect: true,
```

- `keepUnusedDataFor: 0` drops a cache entry the moment its last subscriber
  unmounts, so a remount can never reuse it.
- `refetchOnMountOrArgChange: true` re-issues the request on every mount
  regardless.

Together these mean every arrival on a knowledge-flow-backed page repays the
full network cost. Leaving the Resources page for the chat and coming back
reloads the whole folder tree, the task history and the corpus stats.

`refetchOnFocus` and `refetchOnReconnect` are inert: `setupListeners` is never
called on the store. They should go or be made real, but either way the current
configuration is misleading.

## Amplifier on the Resources page

`TeamResourcesPage` replaces its whole body with a spinner while the health
probe query is in flight, and `getQueryUiState` treats `isFetching` the same as
`isLoading`:

```ts
const kfState = getQueryUiState({ isLoading, isFetching, isUninitialized, isError });
if (kfState === "loading") return <spinner/>;
```

So any refetch of that probe unmounts `DocumentWorkspace`. That discards the
folder the user was in, the loaded document pages and the fetched folder sizes
— and, because of `keepUnusedDataFor: 0`, evicts the cache entries of every
query the workspace owned, which then all refetch on remount. One probe refetch
cascades into a full reload of the page.

Observed in a single session: five requests at T, then the same five again at
T+11s.

## Proposed fix

- give `knowledgeFlowApi` a real `keepUnusedDataFor` and drop the blanket
  `refetchOnMountOrArgChange`, relying on the tag invalidation the comment
  already asks for;
- either call `setupListeners` or remove the two dead refetch flags;
- decouple the page-level spinner from `isFetching` so a background refetch no
  longer unmounts the workspace.

This one touches every page that consumes knowledge-flow, so it wants its own
review pass — hence a separate issue from #2617 and #2618.

Maxime-DARAGON:
Closes #2619.

The knowledge-flow RTK Query slice was configured so that nothing was ever
served from cache. The code already flagged it:

```ts
// todo: remove this to improv app performance. Stale data should be handled with cache invalidation, not with timeouts.
keepUnusedDataFor: 0,
refetchOnMountOrArgChange: true,
```

`keepUnusedDataFor: 0` dropped a cache entry the moment its last subscriber
unmounted; `refetchOnMountOrArgChange: true` re-issued the request anyway. Every
arrival on a knowledge-flow-backed page repaid the full network cost.

## What this does, and what it deliberately does not

The issue originally proposed dropping `refetchOnMountOrArgChange`. On
inspection that would be a regression: **no endpoint in this slice declares
`providesTags` or `invalidatesTags`** — the slice README documents the pattern,
nothing uses it, and `tagTypes` is limited to `Team`/`TeamMember`. Revalidating
at every mount is the only thing that refreshes a list after a mutation, next
to the explicit `refetch()` calls in the pages.

So the mount revalidation stays, and what changes is that the previous answer
is no longer thrown away while it runs:

- `keepUnusedDataFor: 0` → `60`. Freshness semantics are unchanged; a remount
  now renders immediately and refreshes underneath instead of starting from an
  empty loading state.
- The four blob endpoints (`knowledgeFlowApi.blob.ts`) keep a retention of `0`.
  Each resolves to a whole file held in the Redux store, and pinning those for
  a minute after the last viewer unmounted would trade one problem for another.
- `refetchOnFocus` / `refetchOnReconnect` are removed rather than made real:
  `setupListeners` is never called on the store, so they did nothing. Switching
  them on would add traffic nobody asked for, and is a separate decision.

Replacing the mount revalidation with real tag invalidation remains the goal —
it is slice-wide work, not a line in this fix. The comment in the slice now says
so instead of the bare todo.

## The amplifier on the Resources page

`TeamResourcesPage` replaced its whole body with a spinner while the health
probe was in flight, and `getQueryUiState` treated `isFetching` the same as
`isLoading`. So any refetch of that probe unmounted `DocumentWorkspace` — which
owns the folder you are standing in, its loaded document pages and its resolved
folder sizes, none of which survive. And with no subscribers left, every query
that workspace owned lost its cache entry too and refetched on remount. One
probe refetch cascaded into a full reload; a single session showed five requests
at T and the same five again at T+11s.

`getQueryUiState` gains an opt-in `keepPreviousWhileRefetching`, used by that
page's gate. The default is unchanged, so the two other call sites
(`PromptsPage`, `MarketplacePrompts`) behave exactly as before — they drive
inline indicators, which is what the default was written for.

## Tests

- `queryUiState.test.ts` (new) — the option holds `ready` through a refetch,
  still blocks on the very first load, and the default is untouched.
- `TeamResourcesPage.test.tsx` — the workspace's DOM node is the *same node*
  across a background revalidation, which a remount would have replaced.
  Verified to fail without the gate change. Plus the two ends of the gate: the
  page still blocks before the first answer, and still shows the service notice
  on a settled failure.

Full suite: 2078 passed, 2 skipped.

## What is not covered by tests

The retention change itself is configuration; its effect (a remount serving
cached data) is RTK Query behaviour, not ours. Worth a look in the browser
before merge: open a team's Resources page, walk into a folder, navigate to the
chat and back, and confirm the page no longer flashes its spinner.

Part of the same investigation as #2617 and #2618.


dimitri-tombroff:
maxim i let you resolve conflicts on this one

Maxime-DARAGON:
Added after a round of feedback: the return to a knowledge-flow page is now
actually free, and the Resources toolbar carries the control that makes that
acceptable.

`refetchOnMountOrArgChange: true` → `30`. With the retention already in place,
the previous commit made a remount render instantly but still pay the round
trip; walking to the chat and back now costs nothing at all until the answer
has aged past the window.

That window is bounded staleness — a change made from another screen can stay
invisible for it — so `DocumentWorkspace` gains an explicit refresh action, left
of new-folder and upload. One press refetches the folder tree, the open
folder's current page with its offset preserved, the usage stats and the team's
storage quota; the button shows its spinner for the round trip.

It is deliberately **outside** the `canUpdateResources` gate its two neighbours
use: refreshing is not a mutation, and a read-only member needs it as much as
anyone. `DocumentWorkspace.refresh.test.tsx` pins that, along with what one
press actually refetches.

`COMPONENT-UX.md` gains a short section for the new control and the trade-off
behind it. Full suite: 2083 passed, 2 skipped.

PR #2634, opened 2026-09-11
fix(#2633): keep the folder path when removing a document from a tag

### LINKED ISSUE #2633: Resources tab: deleting a document moves its folder to the root
## Symptom

In the Resources tab, deleting a document (single row action or bulk delete) inside a nested folder makes that folder change level in the folder tree: after the refresh it shows up as a root-level folder, and the view being browsed falls back to the root listing.

## Root cause

\`useDocumentCommands.removeFromLibrary\` and \`bulkRemoveFromLibraryForTag\` (apps/frontend/src/components/documents/common/useDocumentCommands.tsx) issue \`PUT /tags/{id}\` with a \`tagUpdate\` body carrying \`name\`, \`description\`, \`type\` and the new \`item_ids\` - but no \`path\`.

Since #2483 the backend applies every editable field from the body (\`tag.path = tag_data.path\` in \`TagService.update_tag_for_user\`), so the missing \`path\` is stored as \`null\` and the folder tag is re-parented to the root. Before #2483 the endpoint ignored those fields, which is why the omission never showed.

\`renameFolder\` already forwards \`path\` and is not affected.

## Fix

Send \`path: tag.path\` in both remove payloads, with a unit test asserting the PUT body preserves the folder path.

simcariou:
Closes #2633

## What

`removeFromLibrary` and `bulkRemoveFromLibraryForTag` (`useDocumentCommands.tsx`) now send `path: tag.path` in the `PUT /tags/{id}` body.

## Why

Both commands rebuild the tag payload from scratch and left `path` out. That was harmless until #2483 made `TagService.update_tag_for_user` apply every editable field of the body: the missing `path` was stored as `null`, so deleting a document inside a nested folder re-parented that folder to the root of the Resources tree, and the view being browsed fell back to the root listing.

`renameFolder` already forwards `path` and is unchanged.

## Tests

- New `useDocumentCommands.remove.test.tsx`: single and bulk removal both assert the PUT body carries the folder's current `path` alongside the trimmed `item_ids`.
- `npx vitest run` on the remove + rename test files: 4 passed. `tsc --noEmit` and prettier clean.

Not verified: manual run in the browser.

### DIFF
--- a/apps/frontend/src/components/documents/common/useDocumentCommands.remove.test.tsx
+++ b/apps/frontend/src/components/documents/common/useDocumentCommands.remove.test.tsx
@@ -1,0 +1,150 @@
+// @vitest-environment happy-dom
+// Copyright Thales 2026
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Regression coverage: removing a document from a nested folder moved the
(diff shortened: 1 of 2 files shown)


simcariou:
Closing as a duplicate of #2635 (same fix for the same bug, opened first with one more test case).

Issue #2561, opened 2026-09-04
Bound OpenFGA concurrency server-side (ReBAC engine / ingestion controller), not just per client


dimitri-tombroff:
## Problem

#2547 fixed the reported bulk-upload symptom (concurrent per-file ReBAC/quota
checks tripping OpenFGA `deadline_exceeded`) client-side in
`DocumentUploadDrawer.tsx`: files are now grouped into batched requests and
run through a bounded (4-way) concurrency pool.

That protects one client path, for one well-behaved client. It does not
protect the backend itself:

- `ingestion_controller.py`'s `upload_documents_sync`/`process_documents_sync`
  place no bound on how many concurrent requests they accept and forward to
  OpenFGA — everything protecting OpenFGA from a burst now lives entirely in
  frontend in-memory state.
- A second browser tab, another client (mobile app, a script hitting the API
  directly), or another team member uploading concurrently gets no
  throttling at all.
- `fred-core`'s `RebacEngine` multi-item APis (`has_permissions`,
  `check_user_team_permissions_or_raise`, etc.) fan out via bare
  `asyncio.gather(...)` with no semaphore bounding concurrent OpenFGA calls,
  even within a single call — so any caller with a large item list can
  reproduce the same failure mode.
- #2257 (folder listing: unbounded per-document OpenFGA checks in
  `filter_readable_document_uids`) is the identical class of problem at a
  different call site — this isn't upload-specific.

## Precedent

`apps/control-plane-backend/control_plane_backend/product/service.py` already
has the pattern for exactly this: `_BULK_DELETE_CONCURRENCY = 5` /
`asyncio.Semaphore(_BULK_DELETE_CONCURRENCY)` bounds a bulk-delete loop where
each item does its own authorization + deletion call.

## Suggested direction (not yet decided)

Add an `asyncio.Semaphore` at the layer that would close this class of bug
everywhere at once, rather than per call site:

- Inside `RebacEngine`'s multi-item methods (or the OpenFGA client wrapper
  itself), so every caller is protected once — the more general fix.
- Or, at minimum, around `ingestion_controller.py`'s per-tag ReBAC check loop
  and `filter_readable_document_uids`, mirroring the control-plane
  `_BULK_DELETE_CONCURRENCY` pattern, as a narrower first step.

Whichever layer is chosen should also give the concurrency budget a source
of truth (e.g. `docs/swift/platform/REBAC.md`) rather than a number picked
[...]

PR #2639, opened 2026-09-11
perf(#2631): fold capability can_use locally instead of one ListObjects per team

### LINKED ISSUE #2631: Admin capabilities list: one sequential ListObjects per team with agent instances
## Problem

`GET /admin/capabilities` (the admin Features page) gets slower as the number of users and teams grows. On a platform with a few hundred users who each created a personal-space agent, the page takes tens of seconds to load.

## Cause

The resting-health column (`compute_capability_impact`, `capabilities/impact.py`) resolves `can_use` per team with one OpenFGA `ListObjects(team:T, can_use, capability)` **per team holding at least one agent instance, awaited sequentially** (`_usable_ids_by_team`).

- Every personal space is a team, so the call count grows with the user base, not only with collaborative teams.
- `can_use` is defined with exclusions (`but not disabled`, `but not personal_block`), so OpenFGA cannot answer a `ListObjects` from an index: it runs a `Check` per candidate capability (roughly every default-on capability, ~87). One call is ~87 server-side checks.

500 personal spaces + 20 teams -> ~520 sequential HTTP calls -> ~45k server-side checks per page load.

The rest of the endpoint is already bounded (#2089, #2181): one cached `Read` per capability for the row data, one DB read, one Keycloak count, de-duplicated pod fetches. `preview_revoke_impact` (platform-wide revoke preview) has the same sequential per-team loop.

## Proposed fix

Derive `can_use` locally from the capability tuples the endpoint already reads.

Each row already fetches, per capability, the complete direct tuple set (`enabled[team]`, `disabled[team]`, `default_on`, `personal_on`, `personal_disabled`) through the 45 s write-invalidated `get_enablement_relations_cached`. `schema.fga`'s `can_use` is a pure function of those five facts plus "is the team a personal space" - the `organization#team` / `organization#personal_team` edges are contextual tuples injected for every team / every personal team, so they are always true locally:

```
inherited = (default_on or (personal and personal_on)) and not (personal and personal_disabled)
can_use   = (team in enabled or inherited) and team not in disabled
```

- `_usable_ids_by_team` becomes: collect the capability ids referenced by instances (selected + template), fetch their cached relation sets (shared with the row builds), fold per team in memory. Zero `ListObjects`. Cost is O(referenced capabilities), independent of the number of teams and users.
- The fold lives in fred-core next to `capability_authz.py` (single copy of the `can_use` semantics), with a truth-table test mirroring `schema.fga`. Same shape as `_fold_team_role_relations` for `team_member`.
- `preview_revoke_impact` reuses the same path.
- No authorization decision is served from this fold: runtime execution and agent-save checks keep asking OpenFGA live. This is an administration display cache in the sense of `REBAC.md`.

## Out of scope (follow-ups)

- `GET /teams/all`, loaded by the same page, does one cached `Read` per collaborative team in an unbounded gather (see #2561).
- The capability materialization admin action (`product/service.py`, `usable_capability_ids` per team in a loop) has the same sequential shape; one-shot, left as is.
- Paginated `Read` page size for capabilities with many explicit per-team grants.

simcariou:
Closes #2631

## Problem

`GET /admin/capabilities` (the admin Features page) got slower as the user base grew. The resting-health column resolved `can_use` with one OpenFGA `ListObjects(team:T, can_use, capability)` per team holding an agent instance, awaited sequentially. Every personal space is a team, and `can_use` has exclusions, so each call cost OpenFGA one `Check` per catalog capability: ~500 personal spaces meant ~45k server-side checks per page load. `preview_revoke_impact` had the same loop.

## Fix

Derive `can_use` locally from the per-capability tuple sets the endpoint already reads for the row data.

- **fred-core**: `CapabilityEnablementFacts.from_relations` folds one capability's direct tuples into the five facts `schema.fga`'s `capability#can_use` depends on; `can_team_use_from_facts` is the formula itself. The `organization#team` / `organization#personal_team` edges are contextual and injected for every team / every personal team, so they are always true locally. Deliberately not exported from the top-level `fred_core` API: it is a display fold, never an admission decision.
- **impact.py**: the health column folds each *referenced* capability from the 45 s write-invalidated cache the row builders share (one cached `Read` per referenced id, concurrent). Verdicts are evaluated per instance dependency, not per (team x id). The revoke preview reads the one capability it is about **fresh** (single uncached `Read`) and answers both the explicit-grant exclusion and the already-broken verdict from that same read: an admin confirms a mutation against that number, so it must not lag behind another replica's write. The grant revive path keeps the live `usable_capability_ids`.
- **service.py** (separate `refactor` commit): the listing rows build their columns from the same `CapabilityEnablementFacts`, so the row data and the health column can no longer drift.
- **REBAC.md**: states which path reads the cache, which reads fresh, and the 45 s cross-replica window.

Cost is now O(referenced capabilities), independent of the number of teams and users. The `rebac_operation=list_objects` timer drops to zero on this path, which gives a direct before/after signal in Grafana.

## Verification

- fred-core: 642 passed; control-plane: 1148 passed (`make test` in both modules). ruff + basedpyright clean on touched files.
- Truth-table unit test with hand-written expected rows for every `can_use` branch (explicit grant, opt-out beats grant, default-on inheritance for collaborative and personal spaces, personal class on/off, explicit grant surviving a personal block, no cross-team leak).
- Integration test `test_local_can_use_fold_agrees_with_openfga` run against a real OpenFGA (`docker-compose.integration.yml`): the fold equals `ListObjects` for a personal and a collaborative team. Mutating the fold to the wrong reading (`default_on` never reaching personal spaces) makes both the unit table and the integration test fail.
- Impact-level tests pin: zero `lookup_resources` calls whatever the team count, one read per referenced id, team-level opt-out on a default-on capability, ReBAC-disabled path, and the preview staying correct with a poisoned cache.
- Reviewed with `/code-review` (8 angles) and the `fred-performance-reviewer` pass; all findings applied except the follow-ups below.

## Follow-ups (not in this PR)

- `enablement.py` still inlines the `inherited` formula on two write paths; consolidating them onto the shared fold is a separate change.
- The capability materialization admin action (`product/service.py`) keeps a sequential per-team `usable_capability_ids` loop (one-shot admin action).
- Raw tuple writers that bypass `invalidate_enablement_relations_cache` (template-id namespace migration, seeding) can leave the health column and rows up to 45 s stale on the same replica.
- Cold-cache fan-out bound / single-flight on `get_enablement_relations_cached`: #2561.
- `GET /teams/all`, loaded by the same page, does one cached `Read` per collaborative team in an unbounded gather.

### DIFF
--- a/apps/control-plane-backend/control_plane_backend/capabilities/impact.py
+++ b/apps/control-plane-backend/control_plane_backend/capabilities/impact.py
@@ -40,19 +40,35 @@ Unreachable pod = UNKNOWN, never "broken": the reconciliation sweep skips such
 instances (`skipped_unreachable`) rather than suspending them on a transient
 outage (#1975, RFC §3.9), and this read reports the same way. Telling an admin
 "12 agents broken" because a pod is restarting would be a lie with consequences.
+
+Where `can_use` comes from: the health column folds it from the cached
+per-capability tuple sets, the revoke preview from one fresh read of the single
+capability it is about. `resolve_availability_for_team` (the grant revive path)
+keeps the live `usable_capability_ids` - it feeds a write, not a display.
 """
 
 from __future__ import annotations
 
+import asyncio
 from dataclasses import dataclass, field
+from typing import Sequence
 
+from fred_core import RebacDisabledResult, RebacEngine
 from fred_core.common import TeamId
+from fred_core.security.rebac.capability_authz import (
+    CapabilityEnablementFacts,
+    can_team_use_from_facts,
+)
 
 from control_plane_backend.agent_instances.store import (
     AgentInstanceRecord,
     AgentInstanceStore,
 )
 from control_plane_backend.capabilities.authz import usable_capability_ids
+from control_plane_backend.capabilities.enablement import (
+    cap_ref,
+    get_enablement_relations_cached,
+)
 from control_plane_backend.product.dependencies import ProductServiceDependencies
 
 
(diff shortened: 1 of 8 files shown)


simcariou:
## Measured on a seeded 1000-team / 1000-user local stack

Seeded into the local docker-compose stack: 1000 Keycloak identities, 1000 collaborative teams with 1300 real membership tuples, and 2000 agent instances spread one per collaborative team and one per personal space, so `compute_capability_impact` sees **2003 distinct team ids** referencing 11 capabilities.

Both shapes measured in one process against the same live OpenFGA, so the only variable is the access pattern. The old shape is still in the codebase (`usable_capability_ids`), so it can be timed directly rather than reconstructed.

| Step | Before | After |
| --- | --- | --- |
| Health column, cold cache | 8.0 s | 20 ms |
| Health column, warm cache | 8.0 s | 9 ms |

The old shape costs 3.6 ms per team and scales linearly with the number of teams holding an agent instance; the new one is 11 gathered reads plus a 7 ms local fold over 2008 instances, independent of the team count.

### Correctness on the same data

For every one of the 2003 teams, the fold's verdict per referenced capability was compared against what OpenFGA's `ListObjects` returns:

```
baseline (inheritance only):   2003 teams x 11 capabilities = 22033 verdicts, 0 mismatches
with per-team opt-outs:        2003 teams x 11 capabilities = 22033 verdicts, 0 mismatches
```

The second pass writes explicit `disabled` tuples on 150 synthetic teams first, so the "most specific wins" branch is covered as well as inheritance; those tuples are removed afterwards. 1001 of the teams are personal spaces, so the personal-class inheritance path is exercised at scale too.

### What the page still costs at this size

| Remaining ReBAC fan-out | Cost |
| --- | --- |
| Row builders, 87 capability reads, gathered | 79 ms cold, 0 ms warm |
| `GET /teams/all`, 1005 team reads, gathered, uncached | 907 ms |

`GET /teams/all` is now the dominant ReBAC cost on that screen. It is loaded by the same page and is listed as a follow-up above, not addressed here.

Issue #2552, opened 2026-09-04
Only the sub-agent invoker closes its event generator; _stream still strands the HITL claim


florian-muller:
## Summary

§8.69 established that an abandoned `_iterate_runtime_event_payloads` generator never
runs the turn's own cleanup, and wrapped the sub-agent invoker's call site in
`contextlib.aclosing` (`agent_app.py:817`, with the rationale in the comment there).

The fix was applied to that one call site. Three others still iterate the generator bare:

| line | function | wrapped? |
|------|----------|----------|
| 818  | `LocalRegistryAgentInvoker.invoke` | ✅ `aclosing` |
| 2982 | `_stream` | ❌ bare `async for` |
| 4814 | `execute` (route) | ❌ bare comprehension |
| 4901 | `evaluate` (route) | ❌ bare comprehension |

`_stream` is the one that matters, because it is the **only** caller where the HITL
resume claim is live. The invoker hard-codes `resume_payload=None` and never sets
`interrupt_id`, so `hitl_claim` is always `None` on a sub-agent turn
(`agent_app.py:3810-3812`). On the `_stream` path both can be set — and
`await hitl_claim.consume()` sits at `:3837`, *inside* the generator, after the loop.

## Failure scenario

A user resumes a HITL-interrupted turn, then closes the tab (or the connection drops)
before the stream finishes. The `async for` in `_stream` is abandoned mid-iteration by
the `CancelledError`, so the generator never resumes past its yield, and three things
never run:

- `await hitl_claim.consume()` — the claim is **not** consumed, so it strands until its
  TTL expires rather than being released at disconnect;
- the root trace span never ends, so the turn appears open in Langfuse;
- `await runtime.dispose()` in the outer `finally` falls back to the GC path.

Before §8.69 all callers were equally GC-timed, so this was uniform. Now the sub-agent
path is deterministic and the primary user-facing path is not — the asymmetry is the
bug, and `_stream` is the wrong side of it.

## Fix

Wrap the remaining three call sites in `contextlib.aclosing`, the same way `:817` does.
The two route comprehensions at `:4814` / `:4901` are lower-stakes (they drain to
completion in the normal case) but there is no reason for them to differ.

Then widen §8.69 to say the guarantee holds for every caller, rather than describing the
invoker alone — as written it reads as a sub-agent property when it should be a property
of the generator's contract.

## Acceptance criteria

- [ ] All four `_iterate_runtime_event_payloads` call sites close the generator deterministically.
[...]

Issue #2544, opened 2026-09-04
Sub-agent HITL gate can send an empty tools array to the provider


florian-muller:
## Summary

`SubAgentHitlMiddleware` drops approval-gated tools before the model sees them, but puts
no floor under the result. When **every** tool bound to the agent is unconditionally
gated, `kept` is empty and the middleware calls `request.override(tools=[])`.

Both Anthropic and OpenAI reject an empty `tools` array, so the depth-≥1 turn fails
outright instead of degrading to a no-tool answer — the opposite of what §8.64 promises
("a sub-agent's HITL gate refuses instead of interrupting").

`libs/fred-runtime/fred_runtime/react/middleware/hitl.py`, `awrap_model_call`:

```python
if len(kept) != len(request.tools):
    request = request.override(tools=kept)
```

## Failure scenario

An agent whose only capability contributes approval-gated tools (a single
`writable-document` write tool, say) has `subagent` enabled. The parent calls
`run_subagent`. At depth 1 the gate strips the one tool, `kept == []`, the override
sends `tools: []`, and the provider 400s. The parent receives a transport error rather
than the child's best-effort answer.

The gap is invisible today because every fixture in the suite binds at least one
ungated tool alongside the gated one.

## Fix

Leave `request.tools` untouched when `kept` is empty — the sub-agent framing already
names the removed tools, so the child is told why it cannot act, and the inherited
`aafter_model` refusal remains the safety net for anything still visible.

## Related gap in the same path

`SUBAGENT_APPROVAL_UNAVAILABLE` (`hitl.py:53`, used at `:503`) appears in **no test** —
`grep` finds only the two source lines. The only coverage is
`test_react_middleware_frame.py` asserting the class is present in the frame. That
leaves #2526's first two acceptance criteria unverified, and `_refuse_without_a_human`
resting on an unproven assumption about LangChain's model→tools edge (that a call
already carrying a `ToolMessage` is skipped).
`test_hitl_resume_langgraph_integration.py` already has the harness to prove it.

## Acceptance criteria

- [ ] A depth-≥1 turn whose tools are all unconditionally gated still reaches the model and answers.
- [ ] A test covers the all-gated case explicitly (no fixture silently binding an ungated tool).
- [ ] `SUBAGENT_APPROVAL_UNAVAILABLE` is exercised end to end, proving the refusal path fires.

Issue #2526, opened 2026-09-03
Sub-agent: children cannot hang on human approval (HITL at depth ≥ 1)


florian-muller:
## Parent

#2520 — design in `docs/swift/rfc/SUBAGENT-CAPABILITY-RFC.md` §5.6 (option B, decided 2026-09-03). Tier 1.

## What to build

A sub-agent can never hang its parent waiting for a human. Without a checkpointer an interrupt has nowhere to persist and no UI channel to reach, so at invocation depth 1 or more the existing HITL middleware does two things: it hides unconditionally approval-gated tools from the model, and any call that would still have interrupted (operator `always_require_tools`, conditional `when` predicates) returns an error tool result saying the tool requires human approval and is unavailable in a sub-agent, instead of calling `interrupt()`.

The middleware already owns the full gated set (capability specs plus the operator list); the `subagent` capability's `tools()` cannot see it, which is why this lives runtime-side and not in a second middleware. Depth reaches the frame builder as one more argument next to the capability HITL bindings.

## Acceptance criteria

- [ ] At depth ≥ 1, unconditionally gated tools are absent from the child's tool schema.
- [ ] At depth ≥ 1, a call that would have interrupted returns the error tool result and the child's turn completes.
- [ ] Depth 0 behaviour is byte-for-byte unchanged (existing HITL tests still pass).
- [ ] The child's framing states generically that approval-gated tools are unavailable; it does not need to name them.
- [ ] `RUNTIME-EXECUTION-CONTRACT.md` §14 entry updated.
- [ ] `make code-quality` + `make test` green; `/code-review` run.

## Blocked by

- #2525 — Sub-agent POC

PR #2644, opened 2026-09-11
docs(HITL-01): propose HITL occurrence identity and human response contract

### LINKED ISSUE #2642: [Epic] HITL expressivity: agent-initiated questions with choice, free text and comment
## Problem

Human-in-the-loop today has exactly one producer on the modern runtime: the tool
approval gate in `FredHitlMiddleware.aafter_model`. It always emits the same frozen
payload - two choices (`proceed`/`cancel`), `free_text=False` - and its answer is read by
`_is_cancelled_human_decision()`, which only distinguishes "cancel" from "not cancel".
Everything else in the resume payload is discarded.

The wire contract is already richer than that. `HumanInputRequest` carries `title`,
`question`, `choices[]`, `free_text` and `metadata`, and `HitlPrompt.tsx` already renders
both choice buttons and a free-text area. `HitlSpec.allowed_decisions` is even documented
as "forward-compat: the gate renders proceed/cancel today; richer decisions are a later,
deliberate contract amendment".

So the gap is not expressiveness of the contract. It is that nothing but the tool
approval gate can ever ask, and nothing consumes an answer richer than cancel.

Arbitrary business questions do exist, but only on the legacy Graph V2 runtime
(`context.request_human_input()` / `choice_step()`), which returns a bare `choice_id` -
and #1080 records that capability HITL is bypassed there.

## Outcome

An agent can ask its user a question at any point during a turn - a single choice among N
options (yes/no included), free text, or a choice plus a free-form comment - and use the
answer to continue the same turn.

---

# Planning record

This section is the complete record of the planning session: every question asked, every
answer given, the options considered and rejected, and the facts established by reading
the code and probing the installed LangGraph. It is written out in full so no decision
has to be reconstructed from memory later.

## Established facts about the current system

Read from the code before any decision was taken:

1. `HumanInputRequest` (`fred-sdk/contracts/runtime.py`) already carries `stage`,
   `title`, `question`, `choices[]` (`HumanChoiceOption`: `id`/`label`/`description`/
   `default`), `free_text`, `metadata`, `checkpoint_id`, `interrupt_id` and
   `pending_calls[]`. Yes/no, single choice and free text are already expressible.
2. `HitlPrompt.tsx` already renders choice buttons and a free-text textarea, and applies
   the composer's `maxChatInputChars` limit to the typed answer.
3. ReAct V2 has exactly one emission point: `FredHitlMiddleware.aafter_model`. It builds
   a fixed payload via `build_tool_approval_request` - `proceed`/`cancel`, and
   `free_text=False` set deliberately, with the comment "a free-text answer here has
   nowhere to go [...] Showing the box anyway silently discarded whatever the human
   typed".
4. The gate already batches: several gated calls in one turn produce a single combined
   interrupt (#2177), because proceed/cancel applies to the whole set atomically.
5. `_is_cancelled_human_decision()` reads only cancel vs not-cancel. The resume payload's
   `answer` and `text` fields reach the backend and are discarded.
6. Gating sources are capability `HitlSpec`s (`require` / `when`) and the operator's
   `ToolApprovalPolicy.always_require_tools` exact list. The old name-prefix heuristics
   were retired in #1978.
7. Graph V2 has arbitrary business HITL through `request_human_input()` and
   `choice_step()`, returning a bare `choice_id`. It is the legacy runtime.
8. History stores `HitlRequestPart` / `HitlResponsePart` in `parts_json` (JSONB), so
   adding fields needs no Alembic migration. `HitlResponsePart` has only `choice_id` and
   `label`; free text is stored by hijacking `choice_id`.
9. `toThreadMessages.ts` reconstructs pending state with `msgs.find(channel ===
   "hitl_request")` - at most one request and one response per exchange.
10. `extract_interrupt_request` in `react_stream_adapter.py` parses any LangGraph
    interrupt payload generically, whatever the call site, so a tool-raised interrupt
    already reaches the client as `awaiting_human` provided its payload is a serialized
    `HumanInputRequest`.
11. The SQL claim table `checkpoint_hitl_claim` is keyed
    `(thread_id, checkpoint_ns, interrupt_id)` and is explicitly "self-inits via
    create_all, deliberately not alembic-tracked". `create_all` creates a missing table
    but never alters an existing one.
12. `hitlRendererRegistry` lets a capability contribute an approval-card renderer keyed
    by gated tool name (WIKI-04).
13. `reasoning_toggle` is the existing precedent for a platform-owned composer control:
    emitted by control-plane under the reserved owner `PLATFORM_CHAT_CONTROL_OWNER =
    "platform"` (reasoning is not a capability), seeded from `params.default`, stored per
    session in `sessionStorage`, and carried to the runtime as `RuntimeContext.reasoning`
    - a tri-state field where absent is explicitly NOT the same as `false`.
14. Deep shares ReAct's `ReActRuntimeToolResolver`, `ReActToolBinder` and
    `build_react_platform_middleware_frame`, and imports `FredHitlMiddleware` directly.
15. Epic #1080 exists and owns the HITL lifecycle (authorization over time, durable
    pauses, discovery, recovery, expiry, audit). It does not cover expressiveness, and
    its design gate has not been passed.

## Probe results

Two throwaway probes were run against the installed LangGraph, not against FRED code, to
settle feasibility before designing.

**Probe 1 - can a tool raise an interrupt, and does the answer come back as a
`ToolMessage`?** Yes, natively. A tool calling `interrupt()` pauses the graph; on resume
the call returns the human's value and the tool's return value lands as a `ToolMessage`
carrying the correct `tool_call_id`. The "agent continues" semantics needed here are free.

**Probe 2 - do several `ask_user` calls in one turn get distinct ids?** No. LangGraph
raises one interrupt at a time and re-interrupts after each resume, but all the pauses
share a single `interrupt_id`:

```
1 ask_user call  -> pause 1: id=f7650104...  q1              distinct ids: 1 of 1
2 ask_user calls -> pause 1: id=b821c649...  q1
                    pause 2: id=b821c649...  q2              distinct ids: 1 of 2
3 ask_user calls -> pause 1: id=d57ce25a...  q1
                    pause 2: id=d57ce25a...  q2
                    pause 3: id=d57ce25a...  q3              distinct ids: 1 of 3
```

`ToolNode` runs all of a turn's tool calls inside a single LangGraph task, and
`Interrupt.id` is a hash of that task's checkpoint namespace. The final message list was
correct in every case - one `ToolMessage` per call, each with its own `tool_call_id`.

**Third fact, from LangGraph's documented semantics and confirmed by the probes:** a
resumed task is replayed from the top. Two consequences were drawn. Anything generated at
pause time - a `uuid4()`, a counter, a timestamp - would differ between the pause and its
replay, so an occurrence identifier must be derived from existing checkpointed data. And
any side effect performed inside a tool *before* its pause would be replayed on resume,
which rules out pausing in the middle of an effectful business tool. A dedicated, pure
`ask_user` tool avoids that trap - one reason the chosen trigger is sound.

### 1. Trigger: a platform tool the model calls

**Decided:** the model asks, through a tool exposed to it.

Rejected: a capability author calling `request_human_input()` from inside a tool
(deterministic, but exposed to the replay-of-side-effects problem above); explicit
workflow steps like today's Graph `choice_step`; widening the operator `ToolApprovalPolicy`
to more gate points.

### 2. Shape: one polymorphic tool

**Decided:** `ask_user(question, choices?, allow_free_text?)` - a single catalog entry
with one description to get right.

Rejected: three specialized tools (`ask_user_confirm` / `ask_user_choice` /
`ask_user_text`) - more directive for the model, but three more catalog entries in every
agent, and "choice plus comment" fits none of them; a two-tool split
(`ask_user_choice` + `ask_user_text`).

### 3. Answer forms

**Decided:** single choice among N (yes/no is the two-option case), free text, and choice
plus free-form comment.

Rejected: multi-select (checkboxes). Supported by neither the contract - `choices` is a
mono-selection button list - nor the frontend, nor history, and not wanted.

### 4. Answer semantics

**Decided:** the answer returns as a `ToolMessage`; the turn continues and the agent
keeps reasoning. Probe 1 shows LangGraph does this natively.

Rejected: turning the answer into a visible user turn in the conversation (more readable
on replay, but pollutes history and blurs who said what); a per-form split between the
two.

### 5. Availability

**Decided:** mounted by default, disabled per conversation from the chat composer,
following the `reasoning_toggle` precedent exactly - a platform-owned chat control, a
`RuntimeContext` field, `sessionStorage` per session.

Rejected: a dedicated capability an admin enables per agent instance (coherent with the
architecture, but an agent would need configuring before it can talk to its own user); an
always-on tool with no off switch.

### 6. Behaviour with no human present

**Decided:** outside an interactive chat the tool is simply not mounted, so the model
cannot call it and nothing can stall. The mechanism is the tri-state already used by
`reasoning`: the composer sends the flag, and an absent flag means "no human", which is
not the same as an explicit `false`.

Rejected: mounting it anyway and returning "no human available" immediately (keeps one
prompt everywhere, at the cost of a wasted tool call); failing the execution closed.

### 7. Escape hatch

**Decided:** a "skip" action that hands control back to the agent with an explicit
`ToolMessage` saying the user did not answer, so a badly posed question cannot strand a
conversation.

Rejected: making an answer mandatory (simplest, but a bad question blocks the
conversation - already a recorded gap in #1080); letting the user answer out of band
through the normal composer; a full turn-abandon path.

### 8. UI

**Decided:** reuse the existing HITL building blocks for visual consistency across the
product; question on top, answer labels bounded to 2-3 lines. The classic HITL UI is
going to be reworked anyway, and those shared blocks will be changed once for every
component that uses them. The card lives in the conversation thread and collapses to a
single line once answered.

Rejected: rendering in the trace like today's approvals (no new rendering work, but a
question addressed to a human buried in a technical fold is easy to miss); a permanently
expanded card in the thread.

### 9. Relationship to the existing approval gate

**Decided:** the two stay distinct - the gate is a platform safety decision, `ask_user`
is a request from the agent - but the gate gains the free-text comment in the same work,
closing its own `free_text=False` note. Rejecting a tool with an explanation the agent can
replan against.

Rejected: unifying both under one mechanism (most aligned with the consolidation phase,
but it refactors a validated security path inside a change that already adds a lot).

### 10. Occurrence identity

**Decided:** discriminate by `tool_call_id`, carried in the prompt and echoed on resume,
with backend validation requiring the pair. It is stable across replay, unique per call,
and the field already exists on `PendingToolCall`.

Rejected: allowing only one `ask_user` per turn (no contract change, but one LLM round
trip per question); merging N questions of one turn into a single multi-question card
(good UX, but that is a form, not a question - a new shape to specify, render and store);
documenting "one at a time" in the tool description without enforcing anything (the
collision stays reachable in production).

### 11. Runtime scope

**Decided:** ReAct V2 and Deep. Deep should inherit the tool through the shared resolver,
binder and middleware frame - to be proven by a test, not assumed. Graph V2 only aligns
`choice_step()` on the enriched response contract and must not regress.

Rejected: exposing `ask_user` to Graph LLM nodes (Graph V2 is the legacy runtime;
investing there runs against the consolidation phase); sub-agents at depth >= 1 (#2526,
#2544, #2552 stay open).

### 12. Prompt-injection guarantees

**Decided:** rely on bounded label length (2-3 lines) and on reusing the shared UI blocks,
because visual consistency across the product was preferred over differentiating this
card. The residual risk is accepted and recorded, not mitigated.

Rejected: marking the card visually as "the agent is asking you", distinct from platform
confirmations; reserving the `proceed`/`cancel` ids so a model-authored question cannot
borrow the gate's success/error styling.

The accepted consequence, stated plainly: until now every HITL question was written by
platform code. With `ask_user` the model writes both the question and the button labels,
in agents that read documents and external sources, so poisoned content can shape a
card that reads like a system request.

### 13. Pause lifetime

**Decided:** out of scope, left with #1080, including any per-turn question cap.

Rejected: a minimal safety net in this work (a maximum number of questions per turn, plus
an abandon path from the UI); handling expiry fully here, which would pre-empt #1080's
unpassed design gate.

Accepted consequence: this work makes that known gap worse. Today a conversation can only
stall behind a gated tool; afterwards any agent can pause at any time, and nothing caps
how many times per turn.

### 14. Delivery shape

**Decided:** three chained OpenSpec changes, narrow on purpose per the consolidation
phase. Rejected: two changes split backend/frontend (neither delivers value alone); a
single end-to-end change (one enormous PR across SDK, runtime, control-plane, history and
frontend).

### 15. Tracking

**Decided:** this umbrella issue, referencing #1080 without being absorbed by it, with one
issue per OpenSpec change beneath it. Rejected: three independent issues with no umbrella;
filing the work as sub-tasks of #1080, where expressiveness would sit behind lifecycle
decisions that do not concern it.

### 16. Naming (proposed defaults, not separately confirmed)

`ask_user` for the tool, `ask_user_toggle` for the composer widget,
`RuntimeContext.ask_user` for the field - chosen to read alongside the existing
`reasoning` / `reasoning_toggle` pair. Open to change while change 2 is being written.

### 17. Milestone

Filed under `swift ga`. `swift-v2.2` was due the day this was created, so the later
milestone was the safer home.

---

## The hard part: occurrence identity

Three existing invariants assume at most one platform-generated HITL pause per exchange.
All three break.

The first is the resume path. It validates `interrupt_id` by exact match against the
checkpoint's pending interrupts, relying on the claim recorded in
`react_stream_adapter.py`: "FRED relies on the narrower fact that `FredHitlMiddleware` has
exactly one `interrupt()` call site, so two DISTINCT FRED HITL occurrences always land in
different tasks and get different ids". Probe 2 invalidates that for a tool-based producer,
and a late answer (double click, second tab, reload) could apply to the wrong question.
The same claim is repeated on the claim table's primary key.

The other two: `toThreadMessages.ts` restores at most one request and one response per
exchange, so a reload would recover only the first question; and `HitlResponsePart` stores
free text in `choice_id`, so there is no sound request/response pairing.

One fix covers all three: `tool_call_id` becomes the occurrence discriminator, in the
request payload, in the resume payload and in history.

## Delivery

**Change 1 - occurrence identity and response contract** (open as #2644, OpenSpec change
`hitl-occurrence-identity`). Carries the occurrence identifier through request, resume and
history; validates the `(interrupt_id, occurrence_id)` pair on resume; scopes the
single-use claim per occurrence without altering the claim table's columns; gives
`HitlResponsePart` a dedicated text field and makes `choice_id` optional; makes history
pairing plural. Every new field optional, so today's flows are unchanged. No Alembic
migration.

**Change 2 - the `ask_user` tool end to end.** The tool and its schema; conditional
mounting driven by the composer flag; the platform chat control and its `RuntimeContext`
field; the skip path and the exact `ToolMessage` wording returned for a choice, a text
answer and a skip; a test proving Deep inherits the tool through the shared frame;
`choice_step()` aligned on the enriched response contract with Graph V2 unregressed.

**Change 3 - HITL UI and gate comment.** The prompt rework (question on top, bounded
labels, collapse to one line after answering), in-thread rendering, and `free_text=True`
on the approval gate so a rejection can carry an explanation the agent replans against.

## Explicitly out of scope

- Pause lifetime, expiry, orphan recovery and any per-turn question cap - see decision 13.
- Sub-agent HITL at depth >= 1: #2526, #2544 and #2552 stay open.
- Multi-select answers - see decision 3.
- Mitigating indirect prompt injection - see decision 12.
- Unifying the approval gate with `ask_user` - see decision 9.
- Exposing `ask_user` to Graph V2 LLM nodes - see decision 11.

## Relationship to #1080

#1080 owns the HITL *lifecycle*: authorization over time, durable pauses, discovery,
recovery, expiry, audit. It does not cover expressiveness, and its design gate has not
been passed. This epic owns expressiveness only and must not be blocked behind lifecycle
decisions that do not concern it.

simcariou:
Planning only - no code is touched. First of three OpenSpec changes under epic #2642,
which lets an agent ask its user a question at any point of a turn instead of only
through the tool approval gate on a few expensive tools.

## Why this one comes first

Resume identity today rests on one narrowly true invariant, recorded in
`react_stream_adapter.py` and again on the SQL claim table: the platform HITL middleware
has exactly one `interrupt()` call site, so two distinct occurrences always land in
different LangGraph tasks and get different ids.

The `ask_user` tool that epic #2642 introduces breaks it. A probe against the installed
LangGraph shows `ToolNode` runs all of a turn's tool calls inside a single task, and
`Interrupt.id` is a hash of that task's checkpoint namespace, so N tool-raised pauses in
one turn share one id:

```
3 ask_user calls in one turn -> pause 1: id=d57ce25a...  q1
                                pause 2: id=d57ce25a...  q2
                                pause 3: id=d57ce25a...  q3
                                distinct ids: 1 of 3
```

Two more places assume at most one platform-generated pause per exchange and break with
it: history pairs a request to a response by finding at most one of each per exchange,
and `HitlResponsePart` stores free text by hijacking `choice_id`.

This change specifies the foundation, ahead of the tool that needs it, so the tool lands
on correct identity rather than retrofitting it afterwards.

## What the change specifies

- An `occurrence_id` naming one pause among those sharing an `interrupt_id`, derived from
  the raising call's `tool_call_id`.
- Resume validation on the `(interrupt_id, occurrence_id)` pair.
- A single-use claim scoped per occurrence.
- A response part storing free text in its own field, with `choice_id` no longer doubling
  as one.
- History pairing many requests and answers within one exchange.

Every new field is optional, so the single-pause flows that exist today are unchanged.

## Two points worth a reviewer's attention

**Why the identifier is derived, not generated.** LangGraph replays a resumed task from
the top, so anything minted at pause time - a `uuid4()`, a counter, a timestamp - would
differ between the pause and its replay and fail its own validation. `tool_call_id` is
assigned by the model call, already checkpointed, and unique per call.

**Why the claim key is composite rather than a new column.** `checkpoint_hitl_claim` is
deliberately not Alembic-tracked: it self-initializes through `create_all`, which creates
a missing table but never alters an existing one. A new primary-key column would need
hand-written DDL on every existing deployment. The alternative is recorded and rejected in
`design.md`.

## Out of scope, on purpose

- The `ask_user` tool, composer control and skip path (change 2); UI rework and free text
  on the approval gate (change 3).
- Pause lifetime, expiry and orphan recovery: those stay with epic #1080, whose design
  gate has not been passed. This work makes that known gap worse and says so - today a
  conversation can only stall behind a gated tool, afterwards any agent can pause at any
  time.
- Sub-agent HITL at depth >= 1 (#2526, #2544, #2552).
- Indirect prompt injection: the model writes both the question and the button labels, and
  the decision was to keep the existing visual building blocks for consistency. The
  consequence is accepted and recorded in `design.md`, not mitigated here.

## Verification

`openspec validate hitl-occurrence-identity --strict` passes; `openspec status` reports
4/4 artifacts complete.

### DIFF
--- a/openspec/changes/hitl-occurrence-identity/.openspec.yaml
+++ b/openspec/changes/hitl-occurrence-identity/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-11
--- a/openspec/changes/hitl-occurrence-identity/design.md
+++ b/openspec/changes/hitl-occurrence-identity/design.md
@@ -51,6 +51,14 @@ one-occurrence-per-task invariant still holds, so `interrupt_id` alone remains s
 Validation therefore requires the pair only when the pending occurrence declares one,
 which is what keeps this change invisible to today's flows.
 
+The same replay rule constrains who may raise a pause at all, which is why the contract
+is worded around a tool call rather than an arbitrary call site: any side effect performed
+inside a tool *before* its pause is replayed when the resumed task re-executes. A pause is
+therefore only safe in a tool that does nothing but ask. That rules out pausing in the
+middle of an effectful business tool, and it is the reason the epic's trigger is a
+dedicated, pure tool rather than a `request_human_input()` an author could call from
+anywhere.
+
 ### Extract pending occurrences, not just pending ids
 
 `_pending_react_v2_interrupt_ids` returns a set of ids read from the pending writes. It
--- a/openspec/changes/hitl-occurrence-identity/design.md
+++ b/openspec/changes/hitl-occurrence-identity/design.md
@@ -1,0 +1,116 @@
+## Context
+
+The resume path validates `request.interrupt_id` against the ids of the checkpoint's
+pending `__interrupt__` writes (`_pending_react_v2_interrupt_ids`), and the SQL claim
+table keys single-use admission on `(thread_id, checkpoint_ns, interrupt_id)`. Both rest
+on the same FRED-specific fact, documented in three places: the platform HITL middleware
+has exactly one `interrupt()` call site, so distinct occurrences always land in distinct
+LangGraph tasks and receive distinct ids.
+
+LangGraph itself does not guarantee that. `Interrupt.id` is a hash of the interrupted
+task's checkpoint namespace, so two `interrupt()` calls in one task share it and are
+matched by call order - already pinned against the installed version by
+`test_langgraph_interrupt_id_semantics.py`. A probe confirmed the consequence for the
+tool the epic introduces: `ToolNode` executes all of a turn's tool calls in one task, so
+N tool-raised pauses in one turn share a single id, and a targeted resume re-interrupts
+on the next sibling rather than distinguishing them.
+
+## Goals / Non-Goals
+
+Goals:
+
+- Make the identity of one pause explicit and verifiable end to end.
+- Keep single-use admission correct when siblings share an `interrupt_id`.
+- Give a human answer a truthful storage shape, and let one exchange hold several.
+- Change nothing observable for the single-pause flows that exist today.
+
+Non-Goals:
+
+- The `ask_user` tool, its composer control, and its skip path (change 2).
+- HITL UI rework and free text on the approval gate (change 3).
+- Pause lifetime, expiry, orphan recovery, and a per-turn question cap: those stay with
+  epic #1080, whose design gate has not been passed.
+- Sub-agent HITL at depth >= 1.
+
+## Decisions
+
+### Derive the occurrence identifier from the tool call, do not generate one
+
+`occurrence_id` is the `tool_call_id` of the call that raised the pause, not a fresh
+identifier minted at pause time.
+
+This is forced by LangGraph's resume semantics rather than chosen for elegance: a resumed
+task is replayed from the top, so anything generated per execution - a `uuid4()`, a
+counter, a timestamp - would differ between the pause and its replay and would fail its
+own validation. `tool_call_id` is assigned by the model call that requested the tool, is
+already checkpointed in the assistant message, and is unique per call within a turn.
+
+The field stays optional. A pause raised outside any tool call - the approval gate, and
+the legacy Graph runtime - has no tool call to derive from, and for those the existing
+one-occurrence-per-task invariant still holds, so `interrupt_id` alone remains sufficient.
+Validation therefore requires the pair only when the pending occurrence declares one,
+which is what keeps this change invisible to today's flows.
+
+### Extract pending occurrences, not just pending ids
+
+`_pending_react_v2_interrupt_ids` returns a set of ids read from the pending writes. It
+becomes a function returning `(interrupt_id, occurrence_id | None)` pairs, read from the
+same writes: the interrupt's payload is the serialized human input request, so the
+occurrence identifier travels with it and needs no second lookup.
+
+The surrounding constraint is preserved: this runs before authorization and before the
+agent definition is resolved, so it must keep reading `CheckpointTuple.pending_writes`
+directly rather than building a compiled graph to call `aget_state`.
+
+### Key the claim by occurrence without altering the claim table
+
+The claim table is deliberately not Alembic-tracked: it self-initializes through
+`create_all`, which creates a missing table but never alters an existing one. Adding a
+primary-key column would therefore need manual DDL on every existing deployment, for a
+table whose whole purpose is short-lived admission bookkeeping.
+
+Instead, the claim's `interrupt_id` column holds the occurrence key: the bare
+`interrupt_id` when no `occurrence_id` exists, and a composite of the two when one does.
+The column is already an opaque string to every query - all of them match it for exact
+equality - so uniqueness per occurrence is obtained with no schema change and no
+migration. The composite form must be unambiguous enough that a bare id can never collide
+with a composite one.
+
+Alternative considered and rejected: add an `occurrence_id` primary-key column. Cleaner to
+read, but it requires hand-written DDL outside the migration system this repository
+otherwise keeps strictly linear, and it buys nothing a composite key does not.
+
+### Store the answer truthfully rather than reinterpreting `choice_id`
+
+`HitlResponsePart.choice_id` currently doubles as the free-text field, by a runtime
+convention inherited from `choice_step`. It gains a sibling text field instead, and
+`choice_id` becomes optional so a pure free-text answer records no choice.
+
+Existing rows are not rewritten and not reinterpreted. A stored `choice_id` keeps
+rendering as it does today; only new rows use the new shape. Message parts are stored as
+JSONB, so this is an additive model change with no migration.
+
+## Risks / Trade-offs
+
+- **A second identity field to keep straight.** The contract already carries
+  `checkpoint_id` (legacy Graph) and `interrupt_id` (ReAct), which have been confused
+  before. `occurrence_id` is a third. Mitigation: it is the only one derived from data the
+  UI already holds, and its docstring states the one rule that matters - it names a pause
+  within an interrupt, never an interrupt and never a checkpoint.
+- **A composite claim key is less legible than a column.** Someone reading the table by
+  hand sees a concatenated value rather than two fields. Accepted deliberately over
+  hand-written DDL on a self-initializing table.
+- **Optional-by-default hides mistakes.** A future pause that should declare an
+  `occurrence_id` but does not would silently fall back to interrupt-only validation.
+  Mitigation: a test asserts that any pause raised from within a tool call declares one.
+- **This change alone is not user-visible.** It ships contract and plumbing whose payoff
+  arrives with change 2. That is the point of sequencing it first, but it means its own
+  verification rests on tests rather than on a demonstrable behaviour change.
+
+## Deferred
+
+- The resume payload's `skipped` form belongs to the `ask_user` tool's semantics and is
+  specified in change 2, not here.
+- Whether a pending-occurrence projection API is needed for reliable multi-tab
+  convergence stays an open question under epic #1080; this change deliberately keeps
+  history as the discovery surface it is today.
--- a/openspec/changes/hitl-occurrence-identity/proposal.md
+++ b/openspec/changes/hitl-occurrence-identity/proposal.md
@@ -1,0 +1,75 @@
+## Why
+
+Human-in-the-loop resume identity rests on one narrowly true invariant, recorded in
+`react_stream_adapter.py` and again on the SQL claim table: "`FredHitlMiddleware` has
+exactly one `interrupt()` call site, so two DISTINCT FRED HITL occurrences always land
+in different tasks and get different ids".
+
+GitHub epic [#2642](https://github.com/ThalesGroup/fred/issues/2642) breaks that
+invariant on purpose. It introduces an `ask_user` tool the model may call several times
+in one turn, and a probe against the installed LangGraph shows that `ToolNode` runs all
+of a turn's tool calls inside a single LangGraph task, so the successive pauses share
+one `interrupt_id`:
+
+```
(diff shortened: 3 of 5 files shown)


dimitri-tombroff:
excellent i review this now

